PR #65, opened 2015-07-14
Fix out of bound write.

yuyichao:
@carnaval We need your ["ccall sanitizer"](https://github.com/JuliaLang/julia/issues/11945#issuecomment-117172936). The GC is too forgiving for that.

### DIFF
--- a/src/program.jl
+++ b/src/program.jl
@@ -130,6 +130,7 @@ let
                 logs[d] = ""
                 continue
             end
+            resize!(log_bytestring, log_len[1])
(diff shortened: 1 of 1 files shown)


yuyichao:
Broken commit was https://github.com/JuliaGPU/OpenCL.jl/commit/cc0d9d76f5561a34aab504c9eb1f4cc15993bd51 so @jakebolewski :P

jakebolewski:
Thanks @yuyichao!

yuyichao:
This (crashing) bug seems to be included in multiple released versions so there should probably be a new release?

PR #59, opened 2014-12-14
[WIP] Julia v0.4 Compatibility 

vchuravy:
I am working on getting OpenCL.jl to run with Julia v0.4 and this branch will track the changes necessary to do so.

Todo:
  - [ ] String -> AbstractString
  - [ ] Nothing -> Void
  - [ ] Seqfault for callbacks https://travis-ci.org/JuliaGPU/OpenCL.jl/jobs/44025578#L319-L330

PR #66, opened 2015-07-14
Fix 0.4-dev compatibility

yuyichao:
Tests passes locally except 4 that are also failing on 0.3 for me.

One segfault observed during the port which should be fixed by #65 .


yuyichao:
P.S. I didn't touch anything in the `example/` in this PR (at least not yet...)

yuyichao:
Everything in this repo should be compatible (without deprecation warnings) with 0.4-dev (except IJulia notebooks). Nightly CI is enabled and both the release and nightly travis tests passed.

I realized that there is https://github.com/JuliaGPU/OpenCL.jl/pull/59 when I'm half way through this PR and decided to finish it.... Everything in that PR should be already included in this one.

Didn't update `README` about `0.4` since I don't since it's a good idea to claim 0.4 support now. (Also because I was too lazy to do that.)

PR #67, opened 2015-07-16
REQUIRE julia 0.3

tkelman:
assuming no one is testing with 0.3 prereleases

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
-julia 0.3-
+julia 0.3
 Compat


jakebolewski:
Thanks, this package is that old and I've never changed it.

tkelman:
Happens. Is Travis running out of memory on release julia?

jakebolewski:
Using travis to test to GPU stuff has been a source of constant problems.  I don't know what is going on but we are probably not freeing resources properly as we rely on finalizers to do memory management which can cause problems with resource oversubscription in OpenCL drivers.

PR #70, opened 2015-08-09
Tweak tests and fix nothing comparisons

yuyichao:
1. Make the test pass on Intel CPU and NVIDIA GPU drivers on Linux

    * GPU only drivers can return `CL_DEVICE_NOT_FOUND` error when asked for a `CL_DEVICE_TYPE_CPU` context

    * The event returned by `clEnqueueMarker` should in principle return an event of status `CL_QUEUED` but due to the asynchronous nature of OpenCL, it is not forbidden to put it in `CL_SUBMITTED` either. (The Intel CPU driver does this)

    * Disable the compilation output check on Intel CPU driver. Somehow it is very verbose and output info messages even if there's no errors or warnings.

2. Use `===` and `!==` for comparing with `nothing`. Ref https://github.com/JuliaLang/julia/pull/12440

The test suite fully pass (on 0.4-dev at least) with Intel CPU driver 2014_R2 and NVIDIA GPU driver 352.30 now.

### DIFF
--- a/src/api.jl
+++ b/src/api.jl
@@ -43,7 +43,7 @@ include("api/opencl_2.0.0.jl")
 
 function parse_version(version_string)
     mg = match(r"^OpenCL ([0-9]+)\.([0-9]+) .*$", version_string)
-    if mg == nothing
+    if mg === nothing
         error("Non conforming version string: $(ver)")
     end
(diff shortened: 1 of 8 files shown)


jakebolewski:
Thanks!

PR #74, opened 2015-10-14
Buffer with dimensions

dfdx:
This PR may bring a breaking change for some packages, so discussion is highly welcome. 

Unlike CUDA which supports 1-, 2- or 3-dimensional arrays, OpenCL operates on flat buffers that don't hold any shape information. This makes implementation of high-level functions (e.g. from BLAS) more verbose and less convenient. This PR changes internal structure of `Buffer` to hold size tuple instead of single length. It doesn't influence any existing functions, but lets other packages to use this metadata.

Summary of changes: 

 * `buff.len::Int` is replaced by `buff.sz::Tuple{Vararg{Int}}`
 * existing constructors are adapted 
 * `length(buff)` is adapted
 * all tests pass on both - Julia 0.3 and 0.4

What may be broken: 

 * code that uses field `.len` directly

Alternatively, we can introduce completely new type `CLArray` based on `Buffer`, but making type for a single purpose of adding new fields looks like overkill for me. 

cc: @jakebolewski

### DIFF
--- a/src/buffer.jl
+++ b/src/buffer.jl
@@ -3,18 +3,19 @@
 type Buffer{T} <: CLMemObject
     valid::Bool
     id::CL_mem
-    len::Int
+    sz::@compat(Tuple{Vararg{Int}})
     mapped::Bool
     hostbuf::Ptr{T}
-
-    function Buffer(mem_id::CL_mem, retain::Bool, len::Integer) #hostbuf
+    
+    @compat function Buffer(mem_id::CL_mem, retain::Bool, sz::Tuple{Vararg{Integer}})       
+        len = prod(sz)
         @assert len > 0
         @assert mem_id != C_NULL
         if retain
             @check api.clRetainMemObject(mem_id)
         end
         nbytes = sizeof(T) * len
-        buff = new(true, mem_id, len, false, C_NULL)
+        buff = new(true, mem_id, sz, false, C_NULL)
         finalizer(buff, mem_obj -> begin
             if !mem_obj.valid
                 throw(CLMemoryError("Attempted to double free OpenCL.Buffer $mem_obj"))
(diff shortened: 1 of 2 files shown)


SimonDanisch:
Thanks for the PR, but I'd personally be against this. We should have some low-level buffer type which represents the hardware, and then much rather wrap it in some other Array type. We had some discussion about this over at JuliaGPU/AbstractGPUArray.jl#2. I must admit I haven't been very helpful since then, because I had no time and no pressuring use case yet.

But that's just my opinion ;)

Best,
Simon

jakebolewski:
Yes I'm with @SimonDanisch on this one.  I would want to keep the concept of CLBuffer as close to the standard as possible so there is not too much of an impedance mismatch when porting over code from other languages or from online examples.

Getting started on GPUArray's would be great, even if they are only a thin abstraction over buffers for now with limited functionality.  I agree that they would make calling CLBlas and CLFFT much more convenient.

src/buffer.jl
```diff
@@ -26,12 +27,20 @@ type Buffer{T} <: CLMemObject
...
+    Buffer{T}(mem_id, retain, (len,))
+end
+
+
 Base.ndims(b::Buffer) = 1
 Base.eltype{T}(b::Buffer{T}) = T
-Base.length{T}(b::Buffer{T}) = @compat Int(b.len)
-Base.sizeof{T}(b::Buffer{T}) = @compat Int(b.len * sizeof(T))
+Base.size{T}(b::Buffer{T}) = b.sz
+Base.length{T}(b::Buffer{T}) = @compat Int(prod(size(b)))
```
jakebolewski: I believe this definition already is in Base :-)

test/test_buffer.jl
```diff
@@ -20,7 +20,7 @@ facts("OpenCL.Buffer") do
                                          length(testarray)) --> not(nothing) "no error"
 
             buf = cl.Buffer(Float32, ctx, cl.CL_MEM_ALLOC_HOST_PTR | cl.CL_MEM_READ_WRITE, length(testarray))
-            @fact buf.len --> length(testarray)
+            @fact length(buf) --> length(testarray)
```
jakebolewski: It would be great to open another PR with these changes to the tests

dfdx:
Closeness of current implementation of `Buffer` to C API was my concern too, so I agree that GPUArrays is much better approach. Since I'm directly interested in this being implemented, I'll invest some time into it. 

One question regarding GPUArrays: I see conflicting definitions (e.g. for `GPUArray` in `core.jl` and `gpu_array.jl`), so I suppose this repo is currently more like a scratch, is it?

jakebolewski:
Yes, this repo is just a prototype at the moment.  I think abstract (buffer type agnostic) GPU array stuff should go into GPUArrays as a common infrastructure / api that OpenCL and Cuda can share.  OpenCL array specific stuff can live in the OpenCL.jl package.

SimonDanisch:
Oh, it definitely needs some rework... It is a prototype that is working well enough for me, so I haven't invested more work into it... The code is quite a mess. I promised myself to clean it up when GLVisualize is released.

dfdx:
> OpenCL array specific stuff can live in the OpenCL.jl package.
Since CUDArt.jl already has `CudaArray`, having `CLArray` in OpenCL.jl looks like the right way to go. 

I'm closing this PR as not relevant any more.

PR #75, opened 2015-10-14
Using length() in tests

dfdx:


jakebolewski:
Hmm.  It seem like the OSX OpenCL runtime does not like using a workgroup size > 1.

jakebolewski:
Thanks!

PR #76, opened 2015-10-16
[WIP] CLArray basic implementation

dfdx:
Basic implementation of multidimensional arrays on top of OpenCL buffers. 

Note 1: this PR includes only main code, but not tests or documentation. They will be added just after array design is agreed.
Note 2: arrays have pretty wide API, current implementation includes only little portion of functions, so others will definitely be added as needed. 

Usage example: 

```
import OpenCL
const cl = OpenCL

device, ctx, queue = cl.create_compute_context()
hostA = rand(100, 50)
A = cl.CLArray(ctx, hostA)
B = A'
hostB = cl.to_host(queue, B)
hostB == hostA'

```


vchuravy:
@SimonDanisch Might that be something that would be a better fit for GPUArray?

dfdx:
@SimonDanisch I believe GPUArrays.jl should be one step higher abstraction. I see at least 2 reasons for this: 

 * CUDArt.jl already provides array type with somewhat different internal structure. We could, of course, try to unify them (which would basically mean unifying CUDA and OpenCL models), but for me it looks much easier to unify them on array level, i.e. have separate `CudaArray` and `CLArray` with the same list of basic operations implemented differently. GPUArray may then use these basic operations to provide higher-level API. 

 * Packages like CUBLAS.jl and CLBLAS.jl require exact array type and not abstract `GPUArray`. I think it's ok to have `CLArray <: GPUArray`, but I see neither possibility, nor need to use `GPUArray` in CLBLAS.jl, for example.

SimonDanisch:
I looked shortly over the code and it's exactly what I would like to have in AbstractGPUArrays, as it's easily implementable for any gpu array.
But to be honest, it's pretty time consuming to do this nicely. If this works for you right now, we should just merge it so that you can work with it.
At some point, someone needs to invest at least a weekend to come up with nice abstractions which work with CUDA, OpenCL and OpenGL. I'm happy to do this at some point, but realistacilly I will need some time to get to it. I wouldn't want to hold you back because of our future plans, without actually helping you to get there.

>Packages like CUBLAS.jl and CLBLAS.jl require exact array type and not abstract GPUArray.

I'm not sure what the problem is here and I find it highly desirable, that they work with it.
It would be a shame, if you need to stop using a nice GPU array type, as soon as you want to use `CLBLAS.jl`.
You can always define this in `CLBLAS.jl`:
```Julia
function some_function{T <: CLArray}(A::T)
    T(some_function(native_mem(A))))
end
some_function{T <: GPUArray}(A::T) = error("$T is not supported to work with CLBLAS")
```
given that you have `CLArray{T, N} <: GPUArray{T, N}`, or probably better `GPUArray{T,N, Buff <: CLBuffer}`.
If `CLBLAS` is well designed, you could restrict this to the one function that actually sends the work to `CLBLAS` and all other functions could just use `GPUArray` for their arguments.

jakebolewski:
I think we need to implement more functionality until we need to implement a common unifying abstraction.  OpenCL and Cuda are similar enough that it should be possible, OpenGL I don't know enough about to comment.

I think at bare minimum the Array abstraction should be sufficiently powerful enough to dispatch to CLBLAS/CuBLAS and CLFFT / CuFFT.  Once we have some partial implementation of these we can look towards unifying the two.

I think a parametric type like `GPUArray{CLBuffer{Float64}, 2}` should be what we strive for, you could then alias this to `CLArray{Float64,2}` to be backwards compatible.

src/array.jl
```diff
@@ -0,0 +1,70 @@
...
+Base.size(A::CLArray) = istransposed(A) ? reverse(A.size) : A.size
+Base.ndims(A::CLArray) = length(size(A))
+Base.length(A::CLArray) = prod(size(A))
+Base.reshape(A::CLArray, dims...) = begin
+    @assert prod(dims) == prod(size(A))
+    A2 = copy(A)
+    A2.size = dims
+    return A2
+end
+Base.transpose{T}(A::CLArray{T,2}) = begin
```
jakebolewski: Why isn't this transposing the elements of the underlying buffer?

src/array.jl
```diff
@@ -0,0 +1,70 @@
...
+
+copy(A::CLArray) = CLArray(A.buffer, A.size, A.t)
+    
+##  convenient constructors
+
+CLArray{T,N}(ctx::Context, hostarray::AbstractArray{T,N}) = 
+    CLArray(ctx, (:rw, :copy), hostarray)
+
+# TODO: OpenCL may have faster equivalent
+Base.zeros(t::Type, ctx::Context, dims...) = CLArray(ctx, zeros(t, dims...))
```
jakebolewski: This array needs to determine if it is on the host or the device, and then call an appropriate kernel.

jakebolewski:
@dfdx I think this is a good first step.  We really need a simple abstraction to provide a better API for calling CLMath libraries, the rest can come later.

dfdx:
> I'm not sure what the problem is here and I find it highly desirable, that they work with it.
> It would be a shame, if you need to stop using a nice GPU array type, as soon as you want to use CLBLAS.jl.

I mean you probably aren't going to pass, for example, `GPUArray{CudaArray}` to `CLBLAS` - CUDA and OpenCL are similar, but not really compatible libraries (at least I haven't heard about mixing the two). Instead, what you may want is to dispatch `GPUArray{CudaArray}` to CUBLAS.jl and `GPUArray{CLArray}` to CLBLAS.jl. 

So I imagined GPUArray.jl to be built on top of CUDArt.jl and OpenCL.jl (and maybe OpenGL.jl, if it makes sense) and have: 

 * type `GPUArray` backed either by `CudaArray` or `CLArray`
 * submodule `GPUArray.BLAS` that dispatches functions either to CUBLAS.jl or CLBLAS.jl
 * submodule `GPUArray.FFT` that dispatches functions either to CUFFT.jl or CLFFT.jl 
 * submodule `GPUArray.RAND` that dispatcges functions either to CURAND.jl or CLRAND.jl (which is also on my TODO list), etc.

src/array.jl
dfdx: It's for efficiency: many linear algebra routines (especially in BLAS) can use transposed matrices as is without actually rearranging elements. This is how it's done in NumPy, and this makes transposing a very lightweight operation. But if you have counterexamples, I'm open to suggestions.

src/array.jl
dfdx: Thanks for the note, I'll update code accordingly.

src/array.jl
jakebolewski: This would be fine for calling BLAS routines, but I don't know how useful it would be beyond that.  (For instance if you had to do an explicit transpose on the device for a user defined kernel).  I don't think it matters in the short term.

jakebolewski:
@dfdx have you seen `https://github.com/JuliaComputing/ArrayFire.jl`?  I'm just wondering because they have the most complete GPUArray implementation that wraps all the existing clmath libraries.  Is there a reason something like ArrayFire would not work for you?

dfdx:
@jakebolewski Yes, I keep and eye on this project, however I have 2 concerns regarding it: 

 * It's not ready yet, not actively developed and depends on `Cxx.jl`, which itself requires custom build of Julia, which makes ArrayFire not really suitable for wide audience. All of these issues will be solved one day, but it may take really long and I don't see how I can speed it up.
 * Though ArrayFire wraps many CUDA / OpenCL libraries, it doesn't seem to expose their API. For example, here's a page for [ArrayFire BLAS-related operations](http://www.arrayfire.com/docs/group__blas__mat.htm) and it lists only 3 functions out of several dozens in BLAS standard. 

In fact, I'm not that much interested in unified CUDA / OpenCL / OpenGL programming as in GPU / CPU programming. Just that I believe following common direction and paying back to the community is a win-win strategy.

src/array.jl
dfdx: The way NumPy handles it is:

    copy(transpose(A))

But now I start thinking it will make development of new kernels much harder since one will need to consider transpose flag every time. So I'll probably roll back the change.

dfdx:
OK, it took some time to figure out how to make it both - convenient and efficient - but now it looks pretty usable (see tests for usage examples). So I propose to merge it here as an experimental feature (with all corresponding warnings and possibly breaking compatibility) and move on to see how it works. Alternatively, I can move this code to AbstarctGPUArray.jl, but restrict support to OpenCL only - supporting CUDA as well looks pretty complicated for me at the moment.

src/array.jl
```diff
@@ -0,0 +1,163 @@
...
+    format(s, name="Tom")  ==> "Hello, Tom"
+"""
+function format(s::AbstractString; vars...)
+    for (k, v) in vars
+        s = replace(s, "%($k)", v)
+    end
+    s
+end
+
+const PROGRAM_TRANSPOSE = readall(Pkg.dir("OpenCL/src/kernels/transpose.cl"))
```
jakebolewski: We probably would want to lazy load this.

src/array.jl
```diff
@@ -0,0 +1,163 @@
...
+    return Kernel(p, kernel_name)
+end
+
+## other array operations
+
+"""Transpose CLMatrix A, write result to a preallicated CLMatrix B"""
+function Base.transpose!(B::CLMatrix{Float32}, A::CLMatrix{Float32};
+                         queue=A.queue, block_size=32)
+    ctx = context(A)
+    kernel = build_kernel(ctx, PROGRAM_TRANSPOSE, "transpose",
```
jakebolewski: We should have a simple cache here that saves the compiled kernel for a given block size.

jakebolewski:
Looks like a good start!  I left some inline comments.  It would be nice to rebase and clean up some of the commits.

dfdx:
@jakebolewski Fixed it, see #83 for cleaned up version.

PR #77, opened 2015-10-20
handle unknown error codes more cleanly

nstiurca:
I ran into a small issue where when trying to use CLBLAS.jl, I wasn't initializing things properly, and CLBLAS returned an error code that is not listed in the `_cl_error_codes` dictionary in `OpenCL.jl/src/error.jl`. This triggered a `KeyError` which was very confusing for a while. Perhaps the constructor of `CLError` should check for unknown error codes, and do something sensible instead of throw a new error, hiding the original source of the problem.

### DIFF
--- a/src/error.jl
+++ b/src/error.jl
@@ -180,7 +180,7 @@ immutable CLError <: Exception
     desc::Symbol
 
     function CLError(c::Integer)
-        @compat new(c, _cl_error_codes[Int(c)])
+        @compat new(c, get(_cl_error_codes, Int(c), :CL_UNKNOWN_ERROR_CODE))
     end
 end
 
@@ -188,9 +188,5 @@ Base.show(io::IO, err::CLError) =
         Base.print(io, "CLError(code=$(err.code), $(err.desc))")
 
 function error_description(err::CLError)
-    try
-        _cl_error_descriptions[err.code]
-    catch
-        return "no description for error $(err.desc)"
-    end
+    get(_cl_error_descriptions, err.code, "no description for error $(err.code)")
 end


src/error.jl
```diff
@@ -180,7 +180,13 @@ immutable CLError <: Exception
     desc::Symbol
 
     function CLError(c::Integer)
-        @compat new(c, _cl_error_codes[Int(c)])
+        desc = ""
+        try
```
vchuravy: Instead of `try ... catch` it might be better to use `get(_cl_error_codes, Int(c), :UNKOWN_ERROR_CODE)

nstiurca:
Ah, sorry definitely should have been a symbol, not string.

As for `try...catch`, that pattern is used later in the same file to get more detailed error descriptions. Perhaps change both to `get(...)`?

nstiurca:
For what it's worth, the commit message for 99a2bcc should have been 'set `CLError.dest` to a `Symbol`, not a `String`; also use `get(...)` instead of `try... catch`' but I guess BASH mangled the backticks.

src/error.jl
```diff
@@ -188,9 +188,5 @@ Base.show(io::IO, err::CLError) =
         Base.print(io, "CLError(code=$(err.code), $(err.desc))")
 
 function error_description(err::CLError)
-    try
-        _cl_error_descriptions[err.code]
-    catch
-        return "no description for error $(err.desc)"
-    end
+    get(_cl_error_descriptions, err.code, "no description for error $(err.desc)")
```
vchuravy: sub `err.desc` with `err.code` in the default

vchuravy:
Looks good, if you could address my last comment and squash the two commits.

nstiurca:
Done. Thanks for the extra pair of eyes. I really need some sleep now...

PR #78, opened 2015-10-20
silently ignore double free of memory

nstiurca:
The other thing that tripped me up is the existence of a `release!(buf::Buffer)` function which causes a `CLMemoryError` to be thrown by the `finalizer` if I call it manually, eg to release some memory before the garbage collector gets to it. All the other `release` methods silently become no-ops in case of double free, so doing the same for `CLMemObject` and `Buffer` would make more sense.

### DIFF
--- a/src/buffer.jl
+++ b/src/buffer.jl
@@ -16,9 +16,6 @@ type Buffer{T} <: CLMemObject
         nbytes = sizeof(T) * len
         buff = new(true, mem_id, len, false, C_NULL)
         finalizer(buff, mem_obj -> begin
-            if !mem_obj.valid
-                throw(CLMemoryError("Attempted to double free OpenCL.Buffer $mem_obj"))
-            end
             release!(mem_obj)
             mem_obj.valid   = false
             mem_obj.mapped  = false
--- a/src/memory.jl
+++ b/src/memory.jl
@@ -19,9 +19,6 @@ Base.sizeof(mem::CLMemObject) = begin
 end
 
 function release!(mem::CLMemObject)
-    if !mem.valid
-        throw(CLMemoryError("attempted to double free mem object $mem"))
-    end
     if mem.id != C_NULL
         @check_release api.clReleaseMemObject(mem.id)
         mem.id = C_NULL


vchuravy:
I agree with you that this is inconsistent and we certainly should adapt this for the finalizer. The finalizer should check whether the mem_obj is valid if if yes free it and otherwise just continue on its merry way.

For `release!` I am not so sure. If I doubly free a buffer something is fishy in my program so I like it as a safeguard. I would actually argue that maybe the other `release!` methods should adopt this. On the other hand ignoring double frees is not incorrect per se.

@jakebolewski Any opinions?

nstiurca:
What does `clReleaseMemObject(NULL)` do? I couldn't find the answer after a quick glance at the spec, but I know that in C, `free(NULL)` is a no-op, and similarly for `delete` in C++. Perhaps instead of our own validity check, we should just trust the OpenCL return code. That would probably be the least surprising to most OpenCL developers.

vchuravy:
See https://www.khronos.org/registry/cl/sdk/1.2/docs/man/xhtml/clReleaseMemObject.html

If `cl_mem memobj` is Null it will throw an error and we can't call `clReleaseMemObject` as often as we want because we might have a reference to it elsewhere.

jakebolewski:
Just thinking aloud, what is a use case where the finalizer can double free
and that is the correct behavior?  I can't think off the top of my head and
I'm definitely on the side that double free'ing an underlying  refcounted
OpenCL object should raise an error.

nstiurca:
@jakebolewski I think my use case--explicitly freeing memory when I choose rather than wait for garbage collection--is pretty compelling. Currently, this triggers a double free and a `CLMemoryError`. In this case, I think the correct behavior is for finalizer to do nothing if it detects already free'd buffer.

Do you have a compelling use-case where throwing an error on double free would actually help a developer and/or end user?

Another avenue we should discuss is doing away with `release!()` (or at least make it clear that users of OpenCL.jl shouldn't call it directly) and instead just use `finalize()`. It appears that repeated `finalize()` calls are no-ops after the first. So for my purposes, I might just replace `release!(buf)` with `finalize(buf)` and be done with it.

jakebolewski:
i wrote this before finalize existed so I think we should  move to
deprecating release! and just use finalize for the API.

vchuravy:
Good idea. I will work on a PR for that.

PR #79, opened 2015-10-20
Fix Travis for Mac

vchuravy:
The code for the example does not check for the correct work_group size.

Example 09 is probably calculating the work_group size wrongly since `global*local` must be `<= 1024`.

I am going to think tomorrow about a way to correct that.

@jakebolewski If you get to this earlier go ahead :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ julia:
 notifications:
   email: false
 before_install:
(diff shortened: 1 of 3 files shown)


vchuravy:
It is not the nicest fix, but the important part of the test suite works.

PR #80, opened 2015-10-21
deprecate release!

vchuravy:
After the conversation in #78 this deprecates `release!`. It some cases I found it cleaner to not have the finalizer as an anonymous closure, but we should probably decide on one style.

PR #82, opened 2015-10-24
fix finalizer missing methods/undefined variables/deprecations

nstiurca:
This fixes some errors introduced in PR #80.

### DIFF
--- a/src/context.jl
+++ b/src/context.jl
@@ -10,9 +10,9 @@ type Context <: CLObject
         ctx = new(ctx_id)
         finalizer(ctx, c -> begin
             retain || _deletecached!(c);
-            if ctx.id != C_NULL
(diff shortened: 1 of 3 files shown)


vchuravy:
Thanks, I am still a bit puzzled (and embarrassed) by how Travis didn't catch this... Especially since all of these are copy and paste mistakes.

jakebolewski:
Hmm, how was this not caught by the tests? @nstiurca thanks for the fix.

nstiurca:
I think maybe the errors happened in the garbage collection thread and
maybe those aren't escalated to the main thread? I don't know much about
how the Julia GC works.

jakebolewski:
Ah yes, all error's in finalizers are swallowed, only the error message is printed.

PR #83, opened 2015-10-27
CLArray basic implementation

dfdx:
Squashed and cleaned up version of #76


src/array.jl
```diff
@@ -0,0 +1,137 @@
...
+
+function to_host{T,N}(A::CLArray{T,N}; queue=A.queue)
+    hA = Array(T, size(A))
+    copy!(queue, hA, buffer(A))
+    return hA
+end
+
+## other array operations
+
+const TRANSPOSE_PROGRAM_PATH = Pkg.dir("OpenCL/src/kernels/transpose.cl")
```
jakebolewski: I wouldn't use Pkg.dir here as it breaks if you want to relocate the package directory (a known problem).  It would be better to use `@__FILE__` and a relative path from this source file.

jakebolewski:
Looks good.  Just one comment and then I think it is G2G.

dfdx:
@jakebolewski Makes sense. I updated the code.

PR #87, opened 2015-11-07
Implements  event_notify using safe operations

### LINKED ISSUE #86: Current Event callback uses undefined behaviour
I learned today [1] that the solution I implemented for user callbacks in [2] is using undefined/illegal behaviour.
I am quite amazed that it worked as long as it did...

````
code_llvm(OpenCL.event_notify, (OpenCL.CL_event, OpenCL.CL_int, Ptr{Void}))
 
%7 = load %jl_value_t*** @jl_pgcstack, align 8
````

I am not quite sure how to solve that problem without reimplementing `Base.SingleAsyncWork` since we can't pass along `evt_id::CL_event, status::CL_int`.

@jakebolewski Any ideas?

[1] https://groups.google.com/forum/#!topic/julia-dev/9J1GYfCyVpE
[2] https://github.com/JuliaGPU/OpenCL.jl/blob/f529546ac3d9aa0517e1d7e4b72856f45284d096/src/event.jl#L97-L106

### PR DESCRIPTION
**Pull request by vchuravy**
Fixes #86.

````julia
julia> using OpenCL

julia> code_llvm(OpenCL.event_notify, (OpenCL.CL_event, OpenCL.CL_int,Ptr{Ptr{Void}}))

define void @julia_event_notify_21517(i8*, i32, i8**) {
top:
  %3 = load i8** %2, align 1
  %4 = getelementptr i8** %2, i64 1
  %5 = load i8** %4, align 1
  %6 = bitcast i8* %5 to i8**
  %7 = getelementptr i8** %2, i64 2
  %8 = load i8** %7, align 1
  %9 = bitcast i8* %8 to i32*
  store i8* %0, i8** %6, align 1
  store i32 %1, i32* %9, align 1
  call void inttoptr (i64 140256579765648 to void (i8*)*)(i8* inreg %3)
  ret void
}
````

### DIFF
--- a/src/event.jl
+++ b/src/event.jl
@@ -94,23 +94,56 @@ Base.getindex(evt::CLEvent, evt_info::Symbol) = info(evt, evt_info)
     end
 end
 
-function event_notify(evt_id::CL_event, status::CL_int, julia_func::Ptr{Void})
-    # Obtain the Function object from the opaque pointer
-    callback = unsafe_pointer_to_objref(julia_func)::Function
-
-    # In order to callback into the Julia thread create an AsyncWork package.
-    cb_packaged = Base.SingleAsyncWork(data -> callback(evt_id, status))
-
-    # Use uv_async_send to notify the main thread
-    ccall(:uv_async_send, Void, (Ptr{Void},), cb_packaged.handle)
+immutable _EventCB
+    handle :: Ptr{Void}
+    evt_id :: CL_event
+    status :: CL_int
 end
 
-const event_notify_ptr = cfunction(event_notify, Void,
-                                   (CL_event, CL_int, Ptr{Void}))
+function event_notify(evt_id::CL_event, status::CL_int, payload::Ptr{Void})
+    ptr = convert(Ptr{_EventCB}, payload)
+    handle = unsafe_load(ptr, 1).handle
+
+    val = _EventCB(handle, evt_id, status)
+    unsafe_store!(ptr, val, 1)
+
+    # Use uv_async_send to notify the main thread
+    ccall(:uv_async_send, Void, (Ptr{Void},), handle)
+    nothing
+end
+
+function preserve_callback(evt :: CLEvent, cb, ptr)
+    evt._cbs[cb] = 0
+    push!(evt._memory, ptr)
+end
 
 
 function add_callback(evt::CLEvent, callback::Function)
-    @check api.clSetEventCallback(evt.id, CL_COMPLETE, event_notify_ptr, callback)
+    event_notify_ptr = cfunction(event_notify, Void,
+                                   (CL_event, CL_int, Ptr{Void}))
+
+    # The uv_callback is going to notify a task that,
+    # then executes the real callback.
+    cond = Condition()
+    cb = Base.SingleAsyncWork(data -> notify(cond))
+
+    # Storing the results of our c_callback needs to be
+    # isbits && isimmutable
+    r_ecb = Ref(_EventCB(cb.handle, 0, 0))
+
+    @check api.clSetEventCallback(evt.id, CL_COMPLETE, event_notify_ptr, r_ecb)
+
+    @async begin
+       try
+         Base.wait(cond)
+         ecb = r_ecb[]
+         callback(ecb.evt_id, ecb.status)
+       catch
+         rethrow()
+       finally
+         Base.close(cb)
+       end
+    end
 end
 
 function wait(evt::CLEvent)
--- a/src/context.jl
+++ b/src/context.jl
@@ -29,16 +29,27 @@ function Base.show(io::IO, ctx::Context)
     print(io, "OpenCL.Context(@$ptr_address on $devs_str)")
 end
 
+immutable _CtxErr
+    handle :: Ptr{Void}
+    err_info :: Ptr{Cchar}
+    priv_info :: Ptr{Void}
+    cb :: Csize_t
+end
+
 function ctx_notify_err(err_info::Ptr{Cchar}, priv_info::Ptr{Void},
                         cb::Csize_t, julia_func::Ptr{Void})
-    err = bytestring(err_info)
-    private  = bytestring(Base.unsafe_convert(Ptr{Cchar}, err_info))
-    callback = unsafe_pointer_to_objref(julia_func)::Function
-    callback(err, private)::Ptr{Void}
+    ptr = convert(Ptr{_CtxErr}, payload)
+    handle = unsafe_load(ptr, 1).handle
+
+    val = _CtxErr(handle, err_info, priv_info, cb)
+    unsafe_store!(ptr, val, 1)
+
+    ccall(:uv_async_send, Void, (Ptr{Void},), handle)
+    nothing
 end
 
 
-const ctx_callback_ptr = cfunction(ctx_notify_err, Ptr{Void},
+const ctx_callback_ptr = cfunction(ctx_notify_err, Void,
                                    (Ptr{Cchar}, Ptr{Void}, Csize_t, Ptr{Void}))
 
 function raise_context_error(error_info, private_info)
@@ -48,7 +59,7 @@ end
 
 function Context(devs::Vector{Device};
                          properties=nothing,
-                         callback::Union{Void,Function}=nothing)
+                         callback::Union{Function, Void} = nothing)
     if isempty(devs)
         ArgumentError("No devices specified for context")
     end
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: src/event.jl [hunk 1]
```diff
@@ -94,23 +94,30 @@ Base.getindex(evt::CLEvent, evt_info::Symbol) = info(evt, evt_info)
...
-const event_notify_ptr = cfunction(event_notify, Void,
+
+function add_callback(evt::CLEvent, callback::Function)
+    event_notify_ptr = cfunction(event_notify, Void,
                                    (CL_event, CL_int, Ptr{Void}))
 
+    evt_id = Ref{CL_event}(0)
+    status = Ref{CL_int}(0)
+    cb = Base.SingleAsyncWork(data -> callback(evt_id[], status[]))
+    ptrs = [cb.handle, Base.unsafe_convert(Ptr, evt_id), Base.unsafe_convert(Ptr, status)]
```
vchuravy: We need to make sure that this doesn't get GC'd so I am contemplating adding an ObjectIdDict to CLEvent. @jakebolewski Ant better Ideas?

vchuravy:
A problem that I see with this implementation is that each callback has only one set of return values, but it is entirely possible that `event_notify` is called multiple times before or during `data -> callback(evt_id[], status[])` is run, thus overwriting the return values.

vchuravy:
This fix only works on Julia v0.4. @jakebolewski Any ideas how to achieve the same with 0.3?

vchuravy:
I am running into a weird seqfault, 

````
julia -i -L test_ev.jl

signal (11): Segmentation fault
jl_field_index at /usr/bin/../lib/julia/libjulia.so (unknown line)
jl_f_get_field at /usr/bin/../lib/julia/libjulia.so (unknown line)
unknown function (ip: 0x7f62ab466459)
jl_apply_generic at /usr/bin/../lib/julia/libjulia.so (unknown line)
unknown function (ip: 0x7f62ab46554c)
unknown function (ip: 0x7f62ab464480)
jl_apply_generic at /usr/bin/../lib/julia/libjulia.so (unknown line)
unknown function (ip: 0x7f62ab463d31)
jl_apply_generic at /usr/bin/../lib/julia/libjulia.so (unknown line)
unknown function (ip: 0x7f62ab45dd49)
unknown function (ip: 0x7f62ab4603e0)
jl_apply_generic at /usr/bin/../lib/julia/libjulia.so (unknown line)
unknown function (ip: 0x7f62ab45a51a)
unknown function (ip: 0x7f62ab45ab46)
jl_apply_generic at /usr/bin/../lib/julia/libjulia.so (unknown line)
unknown function (ip: 0x7f62ab45a0bb)
jl_apply_generic at /usr/bin/../lib/julia/libjulia.so (unknown line)
unknown function (ip: 0x7f62aeaf6e28)
unknown function (ip: 0x7f62aeaf7724)
jl_apply_generic at /usr/bin/../lib/julia/libjulia.so (unknown line)
julia_notify_21330 at  (unknown line)
jlcall_notify_21330 at  (unknown line)
task_done_hook at task.jl:145
jl_apply_generic at /usr/bin/../lib/julia/libjulia.so (unknown line)
unknown function (ip: 0x7f62aeb57756)
unknown function (ip: 0x7f62aeb5836c)
unknown function (ip: (nil))
[1]    9941 segmentation fault (core dumped)  julia -i -L test_ev.jl
````

Where:
````
using OpenCL
const cl = OpenCL

ctx = cl.create_some_context()
q = cl.CmdQueue(ctx)

callback_called = false
function test_callback(evt, status)
  callback_called = true
  println("Test Callback $evt, $status")
end

usr_evt = cl.UserEvent(ctx)
cl.enqueue_wait_for_events(q, usr_evt)
mkr_evt = cl.enqueue_marker(q)

cl.add_callback(mkr_evt, test_callback)

cl.complete(usr_evt)
````

````julia
julia> code_llvm(OpenCL.event_notify, (OpenCL.CL_event, OpenCL.CL_int,Ptr{Void}))

define void @julia_event_notify_21517(i8*, i32, i8*) {
top:
  %3 = bitcast i8* %2 to i8**
  %4 = load i8** %3, align 1
  %5 = getelementptr i8* %2, i64 8
  %6 = bitcast i8* %5 to i8**
  %7 = load i8** %6, align 1
  %8 = bitcast i8* %7 to i8**
  %9 = getelementptr i8* %2, i64 16
  %10 = bitcast i8* %9 to i8**
  %11 = load i8** %10, align 1
  %12 = bitcast i8* %11 to i32*
  store i8* %0, i8** %8, align 1
  store i32 %1, i32* %12, align 1
  call void inttoptr (i64 140438188430736 to void (i8*)*)(i8* inreg %4)
  ret void
}
````
@vtjnash Do you have any insight in what I might be doing wrong/or if this is the right direction/
The issue I am facing is to pass data from the callback back to the original Julia process.

File: src/event.jl [hunk 2]
```diff
@@ -94,23 +101,49 @@ Base.getindex(evt::CLEvent, evt_info::Symbol) = info(evt, evt_info)
...
+    event_notify_ptr = cfunction(event_notify, Void,
+                                   (CL_event, CL_int, Ptr{Void}))
+
+    r_evt_id = Ref{CL_event}(0)
+    p_evt_id = Base.unsafe_convert(Ptr{CL_event}, r_evt_id)
+    r_status = Ref{CL_int}(0)
+    p_status = Base.unsafe_convert(Ptr{CL_int}, r_status)
+
+    cb = Base.SingleAsyncWork(data -> begin
+		println("Received callback")
```
vtjnash: println isn't really allowed here. the SingleAsyncWork callback is being run directly from libuv's callback, so anything that may try to block (such as the flush at the end of printing) could result in a recursive call to the libuv event loop. a better pattern is for the SingleAsyncWork callback to notify a Condition variable, and do the real work on a thread that monitors that variable (a PR to Base to replace the callback with a Condition variable would be a useful improvement to this API).

vchuravy: So instead I would have: 
````
cond = Condition()
cb = Base.SingleAsyncWork(data -> notify(cond))
@async begin
  wait(cond)
  callback(r_evt_id[], r_status[])
end
````

File: src/event.jl [hunk 3]
```diff
@@ -94,23 +101,49 @@ Base.getindex(evt::CLEvent, evt_info::Symbol) = info(evt, evt_info)
...
+end
 
 
 function add_callback(evt::CLEvent, callback::Function)
-    @check api.clSetEventCallback(evt.id, CL_COMPLETE, event_notify_ptr, callback)
+    event_notify_ptr = cfunction(event_notify, Void,
+                                   (CL_event, CL_int, Ptr{Void}))
+
+    r_evt_id = Ref{CL_event}(0)
+    p_evt_id = Base.unsafe_convert(Ptr{CL_event}, r_evt_id)
```
vchuravy: @vtjnash Is this the correct way to obtain the pointer to the memory that a `Ref` is holding?

src/event.jl
vchuravy: Could it be that the main issue I am running into is that I am trying to write memory in a protected address space?

vchuravy:
@vtjnash Could you take another look?

This is my most promising attempt yet and the callback compiles to:
````
julia> code_llvm(OpenCL.event_notify, (OpenCL.CL_event, OpenCL.CL_int,Ptr{Void}))

define void @julia_event_notify_21521(i8*, i32, i8*) {
top:
  %3 = bitcast i8* %2 to %_EventCB*
  %4 = bitcast i8* %2 to i8**
  %5 = load i8** %4, align 8
  %6 = insertvalue %_EventCB undef, i8* %5, 0
  %7 = insertvalue %_EventCB %6, i8* %0, 1
  %8 = insertvalue %_EventCB %7, i32 %1, 2
  store %_EventCB %8, %_EventCB* %3, align 1
  call void inttoptr (i64 139700456842640 to void (i8*)*)(i8* inreg %5)
  ret void
}
````

jakebolewski:
I believe this solution should work, although should there be a corresponding `unpreserve_callback` otherwise all registered callbacks would never get GC'd.

vchuravy:
@jakebolewski Thanks. Before this goes in we would need to bump the minimal version to 0.4 so I will start a PR for that.

jakebolewski:
I think dropping support for 0.4 is reasonable.  Some parts of the API can be simplified with call overloading.

vchuravy:
I hope you meant dropping support for 0.3 ;)

vchuravy:
https://github.com/JuliaGPU/OpenCL.jl/blob/master/src/context.jl#L32-L38 Should probably receive the same treatment?

vchuravy:
So rebased and squashed and I will start to fix the callback in context.jl, because baased on the docs [1] the callback can/will happen asynchronously.

[1] https://www.khronos.org/registry/cl/sdk/1.0/docs/man/xhtml/clCreateContext.html

vchuravy:
@jakebolewski Take a look at this and if you agree we can merge, after fixing the oom travis error we can go ahead and tag a new release.

Issue #84, opened 2015-10-27
Documentation


jakebolewski:
Now that we have a shiny new documentation system.

vchuravy:
That would also mean dropping support for 0.3, but I am fine with that.

PR #89, opened 2015-11-10
drop support for v0.3 and remove Compat

vchuravy:
see #87 and  #84


vchuravy:
The travis failure is due to oom of CLArray

Issue #92, opened 2015-11-17
`payload` not defined in `ctx_notify_err()`


dfdx:
Probably just a typo, but still better to track it somewhere until somebody finds time to fix it. 

From [build](https://travis-ci.org/JuliaGPU/OpenCL.jl/jobs/91679929) on Travis (Julia 0.4 on Mac):

```
Error :: (line:505) :: OpenCL.CLArray core functions
      Expression: cl.to_host(A') --> cl.to_host(B)
      UndefVarError: payload not defined
       in ctx_notify_err at /Users/travis/.julia/v0.4/OpenCL/src/context.jl:41
       in clEnqueueNDRangeKernel at /Users/travis/.julia/v0.4/OpenCL/src/api.jl:22
       [inlined code] from /Users/travis/.julia/v0.4/OpenCL/src/macros.jl:4
       in enqueue_kernel at /Users/travis/.julia/v0.4/OpenCL/src/kernel.jl:191
       in transpose! at /Users/travis/.julia/v0.4/OpenCL/src/array.jl:106
       in ctranspose at operators.jl:159
       in anonymous at /Users/travis/.julia/v0.4/FactCheck/src/FactCheck.jl:271
       in do_fact at /Users/travis/.julia/v0.4/FactCheck/src/FactCheck.jl:333
       [inlined code] from /Users/travis/.julia/v0.4/FactCheck/src/FactCheck.jl:271
       in anonymous at /Users/travis/.julia/v0.4/OpenCL/test/test_array.jl:41
       in context at /Users/travis/.julia/v0.4/FactCheck/src/FactCheck.jl:474
       in anonymous at /Users/travis/.julia/v0.4/OpenCL/test/test_array.jl:40
       in facts at /Users/travis/.julia/v0.4/FactCheck/src/FactCheck.jl:448
       in include at /Users/travis/julia/lib/julia/sys.dylib
       in include_from_node1 at /Users/travis/julia/lib/julia/sys.dylib
       in include at /Users/travis/julia/lib/julia/sys.dylib
       in include_from_node1 at /Users/travis/julia/lib/julia/sys.dylib
       in process_options at /Users/travis/julia/lib/julia/sys.dylib
       in _start at /Users/travis/julia/lib/julia/sys.dylib
```
Which leads us to [this line](https://github.com/JuliaGPU/OpenCL.jl/blob/master/src/context.jl#L41).

vchuravy:
Thanks :+1:

PR #93, opened 2015-11-17
Fixed OOM on Travis

dfdx:
This PR fixes OOM on Travis (though tests on Mac still fail because of #92, but it seems to be unrelated to the specified issue).

It's interesting to know the reason of OOM. It turns out on some platforms (notably, Julia 0.4 on Linux) creating new queues costs quite some memory on host. Tests, being independent pieces of code, created new queues every time, which eventually resulted in out of host memory error. This fix simply decreases number of newly-created queues, both in test and main code. 

To prevent such situations it might be useful to introduce a kind of default queue that would be used in case no explicit queue is provided. Details of such default queue aren't clear for me yet, so discussion is highly appreciated.

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -12,9 +12,9 @@ typealias CLVector{T} CLArray{T,1}
 ##  constructors
 
 function CLArray{T,N}(ctx::Context,
-                              queue::CmdQueue,
-                              flags::Tuple{Vararg{Symbol}},
-                              hostarray::AbstractArray{T,N})
+                      queue::CmdQueue,
+                      flags::Tuple{Vararg{Symbol}},
+                      hostarray::AbstractArray{T,N})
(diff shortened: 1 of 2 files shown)


vchuravy:
I rebased you branch onto master (I fixed the callback problem), but it still seems to throw a context error on Mac.

dfdx:
Seem like Mac OS X won't allow us to test `transpose()` properly, since `MAX_WORK_GROUP_SIZE` there is equal to `(1024, 1, 1)`, while `transpose()` requires second dimension to be higher than 1. So I suppressed this test on Mac in Travis tests (on Linux and locally they are still being run) and it fixed tests.

vchuravy:
Thanks for looking into this and :tada: for making Travis green again.

PR #95, opened 2015-11-22
Default queue for context

dfdx:
Quite often we need to work with both - context and queue. Passing both of these variables everywhere looks like an overkill, especially in cases when there's 1:1 correspondence between them. This PR creates a cache for default queues: instead of passing 2 variables or creating new queue from context every time, one can now call `default_queue(ctx)` to get cached value. 

In the PR I also use this capability to provide convenient yet flexible interface for constructing `CLArray`s - if user doesn't pass a queue, the one associated with array's context will be used.

### DIFF
--- a/src/util.jl
+++ b/src/util.jl
@@ -1,8 +1,21 @@
+# cached queues
+const DEFAULT_QUEUES = WeakKeyDict{Context, CmdQueue}()
+
 function create_compute_context()
     ctx    = create_some_context()
     device = first(devices(ctx))
     queue  = CmdQueue(ctx)
+    DEFAULT_QUEUES[ctx] = queue
     return (device, ctx, queue)
+end
+
+function default_queue(ctx::Context)
+    # for some reason WeakKeyDict requires wrapping keys of type Context
+    # into WeakRef(), though other types seem to behave normally
+    if !haskey(DEFAULT_QUEUES, WeakRef(ctx))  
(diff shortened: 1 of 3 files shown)


src/util.jl
```diff
@@ -1,10 +1,21 @@
+# cached queues
+const DEFAULT_QUEUES = Dict{Context, CmdQueue}()
```
vchuravy: This will prevent garbage collection for Contexts, so this should probably be a `WeakKeyDict`

dfdx: Fixed.

dfdx:
Juts noticed that some tests are failing, working on it.

dfdx:
Not sure why, but using `Context` as a key in `WeakKeyDict` requires manual wrapping it into `WeakRef`. Anyway, with latest commit this works again, so all tests pass.

jakebolewski:
Hmm, I see the appeal of this api change but having global variables generally make me skeptical that this is the only / best way of doing it.  Why not have a type that couples the context and queue (`CtxQueue`) and then parameterize methods to use that?

dfdx:
I just understood that it's possible to retrieve `Context` from `CmdQueue` via `cl.info(queue, :context)`, so looks like the best approach would be to pass queue everywhere. I need to think about it for a while. 

Thanks for comments. I appreciate filtering ideas through collaboration.

dfdx:
I tried this idea on arrays, and code looks really nice now. Thus I'm closing this PR in favor of #96

PR #96, opened 2015-11-25
Retrieving context from queue, array functions simplified

dfdx:
Since it's possible to retrieve `Context` from `CmdQueue`, there's no need to pass them both everywhere. This PR simplifies array implementation according to this possibility.


jakebolewski:
:+1:, much cleaner!

PR #97, opened 2015-12-04
[WIP] CLArray <: AbstractArray

dfdx:
One little, but important part of array interface.

PR #98, opened 2015-12-07
make CLArray <: AbstractArray

nstiurca:
Based on @dfdx's idea in PR #97, but the tests pass now (at least locally on my machine).

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -1,5 +1,5 @@
 
-type CLArray{T,N} <: CLObject
+type CLArray{T,N} <: AbstractArray{T,N}
     ctx::Context
     queue::CmdQueue
     buffer::Buffer{T}
@@ -32,6 +32,8 @@ end
 
 Base.copy(A::CLArray; ctx=A.ctx, queue=A.queue, buffer=A.buffer, size=A.size) =
     CLArray(ctx, queue, buffer, size)
+Base.copy!{T}(dest::Array{T}, src::CLArray{T}; queue=src.queue)  = copy!(queue, dest, src.buffer)
+Base.copy!{T}(dest::CLArray{T}, src::Array{T}; queue=dest.queue) = copy!(queue, dest.buffer, src)
 function Base.deepcopy{T,N}(A::CLArray{T,N})
     new_buf = Buffer(T, A.ctx, prod(A.size))
     copy!(A.queue, new_buf, A.buffer)
@@ -71,10 +73,11 @@ Base.ndims(A::CLArray) = length(size(A))
 Base.length(A::CLArray) = prod(size(A))
 Base.(:(==))(A:: CLArray, B:: CLArray) =
     buffer(A) == buffer(B) && size(A) == size(B)
-Base.reshape(A::CLArray, dims...) = begin
+Base.reshape(A::CLArray, dims::Tuple{Vararg{Int}}) = begin
     @assert prod(dims) == prod(size(A))
     return copy(A, size=dims)
 end
+Base.reshape(A::CLArray, dims::Int...) = reshape(A, dims)
 
 ##  show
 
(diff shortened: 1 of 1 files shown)


nstiurca:
Hmm, it looks like there might be an issue with Travis on Linux? I don't think it's anything to do with my PR though since the script is failing on `apt-get`.

vchuravy:
I will look into the travis issue later today/tomorrow.

vchuravy:
Can you rebase after #99 has gone through?

jakebolewski:
Although this is desirable, it seems like there are too many performance pitfalls for `AbstractArray` fallbacks in base.  Have you not run into these issues?

vchuravy:
Additionally we then get to the point where you would want to use CLBLAS to power a lot of these algorithms, but we would not want to depend on it in OpenCL.jl. Maybe that is another argument for moving this to GPUArray (or creating GPUArray.jl)?

cc: @SimonDanisch

nstiurca:
#99 seems to have helped, although there are a lot of the following in the Travis log:

    FATAL: Could not load /lib/modules/2.6.32-042stab108.8/modules.dep: No such file or directory
    Error! Fail to load fglrx kernel module! Maybe you can switch to root user to load kernel module directly

That error keeps popping up with every script in `examples/`, but it looks like they still execute/work correctly.

nstiurca:
As for CLBLAS.jl and performance, I'm still trying to get things to work as opposed to get them to work fast, so I really can't say. I'm actually having a lot of trouble with the underlying `libclBLAS`, so I my need to take up some issues with that package before I get things running nicely in Julia for myself.

For now, I'm happy to have a working `CLArray` even if performance isn't ideal yet.

dfdx:
> Although this is desirable, it seems like there are too many performance pitfalls for AbstractArray fallbacks in base. Have you not run into these issues?

Honestly, I don't think we should worry about performance of GPU-based arrays for non-specific operations - having 2 arrays implementing the same interface doesn't imply that they will be fast for _all_ algorithm (e.g. consider difference between dense and sparse arrays and how it affects performance in different use cases).

dfdx:
> Maybe that is another argument for moving this to GPUArray (or creating GPUArray.jl)?

The argument about dependencies is quite valid, but for me it sounds like a question of expectations. PyOpenCL, Boost.Compute, rust-opencl - all of them have array implementations out of the box (as well as CUDArt.jl has `CudaArray`), so if I were new to the topic, I would _expect_ to find arrays in OpenCL itself. 

Also, I still don't see an easy way to create generic `GPUArray`, covering both - OpenCL and CUDA technologies - and I don't see direct profit for this activity. So it will be quite hard to push it into mine (or someone else's) TODO list.

SimonDanisch:
> I would expect to find arrays in OpenCL itself.

Good point, but easily solved by letting OpenCL re-export the GPUArray types.

The biggest benefit of a general GPUArray interface is that you can write higher level libraries which work for both CUDA/OpenCL or maybe even OpenGL arrays.

> covering both - OpenCL and CUDA technologies

I must admit that there are quite a few pitfalls.
I guess most of the functonality must be implemented in `OpenCL.jl`/`CUDA.jl` in the end, but GPUArrays could support things like boundschecking, memory management, `splice!`, `push!` and a richer indexing API.
If `GPUArrays` would force `OpenCL.jl`/`CUDA.jl` to implement a more similar interface, we'd already
made good step forward (e.g. `free`, minimal inteface for buffer types, etc).

dfdx:
> Good point, but easily solved by letting OpenCL re-export the GPUArray types.

Wouldn't it introduce circular dependency? If not, could you please provide a scratch of `GPUArray` implementation?

> I guess most of the functonality must be implemented in `OpenCL.jl`/`CUDA.jl`

Do you really mean `CUDA.jl` and not `CUDArt.jl`? Seems like the latter is much better adopted (e.g. see `CUFFT.jl`, `CUBLAS.jl`, `CUDNN.jl`, etc.)

SimonDanisch:
I guess I mean `CUDArt.jl` then ;)
I can see if I can clean up `AbstractGPUArrays`.
I've actually implemented a map for OpenGL Arrays lately, so it could be a good point in time.

dfdx:
@SimonDanisch Any news on `AbstractGPUArrays`? If not, can we see a map for it?

SimonDanisch:
Oh... Yeah, I'm currently stuffed with work and prioritize releasing GLVisualize.
I see what I can do tomorrow!

dfdx:
Actually no hurry here, I just wanted to refresh the topic to keep it alive, so please take your time.

PR #99, opened 2015-12-07
Fixes travis by enabling restricted and updates repository

vchuravy:
fglrx is only available in restricted and we probably want a more up to date version.

I am hoping that at some point we can switch to the docker infrastructure, but that is depending on:

https://github.com/travis-ci/apt-package-whitelist/pull/750


vchuravy:
The travis failure looks like a network issue.

PR #101, opened 2015-12-08
Correct handling of block size in array transpose

dfdx:
To be fast, matrix transpose on GPU should be done in blocks. Previously, block size was fixed, which resulted in multiple errors. This PR adds computation of maximum allowed block size based on: 

 * maximum dimensions of work-group
 * maximum total size of work-group
 * height and width of a matrix

### DIFF
--- a/src/array.jl
+++ b/src/array.jl
@@ -9,7 +9,7 @@ end
 typealias CLMatrix{T} CLArray{T,2}
 typealias CLVector{T} CLArray{T,1}
 
-##  constructors
+## constructors
 
 function CLArray{T}(buf::Buffer{T}, queue::CmdQueue, sz::Tuple{Vararg{Int}})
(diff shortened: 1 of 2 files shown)


vchuravy:
Particularly nice to see that this fixed the Mac tests.

jakebolewski:
Nice!  Yes, the transpose example in the notebook is definitely not correct for all problem sizes.  @dfdx it would be great if you could add to the notebook showing the performance of `CLArray` transpose in comparison.

dfdx:
@jakebolewski Do you mean comparison of transpose for arrays of different sizes or comparison with transpose for ordinary Julia arrays? Currently I see that OpenCL-based transpose is at least 4 times _slower_. Not sure if it's expected or just my setup, but putting it into notebook as is wouldn't be very impressive :)

PR #104, opened 2016-01-14
Make use of the new travis docker infrastructure

vchuravy:
So the aim of this is to enable the docker infrastructure for travis testing. The problem is that one has to install a working OpenCL implementation manually.

I got the AMD APP SDK working, which is quite nice because we now can test OpenCL 2.0 on travis. The only show stopper right now is a seqfault in waitforevents

https://travis-ci.org/JuliaGPU/OpenCL.jl/jobs/102318248

@dfdx If I can fix the seqfault this should be quite nice, additionally we could think about installing pocl and testing our code against two implementations. I moved back to AMD because one can't install the Intel one without root rights.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: julia
+sudo: false
 
 os:
   - linux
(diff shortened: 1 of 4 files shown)


vchuravy:
This is slightly frustrating. On my local computer I was able to reproduce the issue, but sometimes it is also just stalling (like the last travis build). 

Also the tests are stalling on my local computer with the Intel SDK so it might be a mire general problem.

dfdx:
I'll check if I can reproduce stalling on tests with Intel SDK in my machine.

dfdx:
Hmm, I can't find `libOpenCL.so` after installing Intel SDK. How do you switch between different implementations? 

With AMD and NVidia your branch works fine on my laptop.

vchuravy:
are you using AMD CPU or GPU? Handling the Intel SDK is a bit tricky, I will have to check once I am at work how I have setup things. Also sometimes the error occurs only during GC runs, so it might require several runs too hit it.

vchuravy:
If you are using the Intel sdk you will also have to install the runtime, which apparently contains the .so

vchuravy:
I should tested it on my home computer and we pass all tests on `2014_R2`. The library itself should be in `/opt/intel/opencl-1.2-4.4.0.117/lib64/libOpenCL.so`.

I will need to double check that I wasn't running the tests at work against beignet.

vchuravy:
using an older AMD SDK `2.9.1` seems to fix the test problems.

PR #134, opened 2017-03-22
fix CLArray eltype, better build errors

SimonDanisch:
I found a few more inconsistencies, but I think I will handle them in GPUArrays!

### DIFF
--- a/src/program.jl
+++ b/src/program.jl
@@ -76,17 +76,29 @@ function Program(ctx::Context; source=nothing, binaries=nothing)
     end
 end
 
+function print_with_linenumbers(text, pad = "", io = STDOUT)
+    for (i,line) in enumerate(split(text, "\n"))
+        println(io, @sprintf("%s%-4d: %s", pad, i, line))
+    end
+end
+
 #TODO: build callback...
 function build!(p::Program; options = "", raise = true)
     opts = String(options)
     ndevices = 0
     device_ids = C_NULL
-    api.clBuildProgram(p.id, cl_uint(ndevices), device_ids, opts, C_NULL, C_NULL)
-    if raise
-        for (dev, status) in cl.info(p, :build_status)
-            if status == cl.CL_BUILD_ERROR
-                error(cl.info(p, :build_log)[dev])
-            end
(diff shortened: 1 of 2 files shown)


src/program.jl
```diff
@@ -77,16 +77,15 @@ function Program(ctx::Context; source=nothing, binaries=nothing)
 end
 
 #TODO: build callback...
-function build!(p::Program; options="", raise=true)
+function build!(p::Program; options = "", raise = true)
     opts = String(options)
     ndevices = 0
     device_ids = C_NULL
-    @check api.clBuildProgram(p.id, cl_uint(ndevices), device_ids, opts, C_NULL, C_NULL)
```
vchuravy: Why the removal of `@check`?

SimonDanisch: because it throws, when the compilation fails :( we'd need a more fine grained control, but as it only checks if compilation succeeds and we check it with `info`, we might be fine here!

vchuravy: Yeah then you will need to check the error explicitly

```
err =api.clBuildProgram
if err == THE_ONE_ERROR_WE_CARE_ABOUTE
# custom error handling
elseif err != -1 
   throw
end
```

dfdx:
> With the existence of GPUArrays I am less convinced of the need for CLArray.

Raison d'etre for `CLArray` was supporting `CLBLAS`, but since we are going to cover it in [another place](https://github.com/JuliaGPU/GPUArrays.jl/blob/sd/glsl/src/blas.jl), I believe we can deprecate `CLArray` at the next big release.

SimonDanisch:
Okay, program build failure now always prints the build failure like this:
```
Couldn't compile kernel: 
    1   : void test(int[2] program){
    2   : 
    3   : }
    4   : 
With following build error:
stringInput.cl:1:17: error: expected ')'
stringInput.cl:1:9: note: to match this '('
stringInput.cl:1:13: error: parameter name omitted
```
Then if raise = true, it throws the error from `@check` afterwards.
I purposely don't put the log into the error, since it becomes super hard to read (especially in Atom).
If the error is anything but a build failure, it will just throw immediately!

PR #135, opened 2017-03-23
more types for setarg

SimonDanisch:
This allows more and is potentially more unsafe. 
We could warn when an unknown type is used, or we could read through julia's and opencl padding and mem layout and see if we can come up with better rules!

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -73,7 +73,8 @@ end
 
 function set_arg!(k::Kernel, idx::Integer, arg::CLMemObject)
     @assert idx > 0
-    @check api.clSetKernelArg(k.id, cl_uint(idx-1), sizeof(CL_mem), [arg.id])
+    arg_boxed = Ref(arg.id)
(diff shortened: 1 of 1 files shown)


SimonDanisch:
Lol okay... why is `Ref(C_NULL) === C_NULL` ? We had the same problem in `CUFFT` as far as I remember. This is a serious trap, since generic code that uses `Ref(x)` changes behaviour completely because of that... @yuyichao

yuyichao:
Because Ptr is a Ref. Note that Ref(a) isn't generic anyway. You should either use RefValue or Ref{typeof (x)}

SimonDanisch:
So Ref is actually not a 0 dimensional container, as described in most of
the docs?!
Directly from the Ref docs:

An object that safely references data of type T

Now, if you take the usual constructor semantic, which takes the T from
it's arguments and is supposed to return an instance of Ref, the current
behaviour is just plain confusing... I could get behind convert(Ref,
C_NULL) === C_NULL, since it's not the constructor of Ref!

PR #139, opened 2017-07-26
bug fixes and better aligning

SimonDanisch:


src/context.jl
```diff
@@ -18,6 +16,23 @@ type Context <: CLObject
         return ctx
     end
 end
+# OpenCL gives you the same pointer when creating two context from the same
+# device while not freeing any of them. So we can get two different context with
+# the same pointer, which will not work in the way our finalizer works!
+const _context_cache = Dict{CL_context, WeakRef}()
```
vchuravy: ... Yeah.

src/kernel.jl
```diff
@@ -84,14 +84,148 @@ function set_arg!(k::Kernel, idx::Integer, arg::LocalMem)
...
+        offset[] += sizeof(x)
+    end
+    return
+end
+
+
+
 #TODO: type safe calling of set args for kernel (with clang)
+# is 1024 a good number?
+const _arg_tmp_buffer = Vector{Int8}(1024)
```
vchuravy: This might be racy at some point.

src/kernel.jl
```diff
@@ -84,14 +84,148 @@ function set_arg!(k::Kernel, idx::Integer, arg::LocalMem)
...
     end
-    boxed_arg = Ref{T}(arg)
-    @check api.clSetKernelArg(k.id, cl_uint(idx - 1), sizeof(T), boxed_arg)
+    aligned_arg, off = aligned_convert(arg)
+    if sizeof(aligned_arg) > 1024
+        error("Can't upload single types to opencl that are bigger than 1024 bytes")
+    end
+    T_aligned = typeof(aligned_arg)
+    # is this save?
+    unsafe_store!(Ptr{T_aligned}(pointer(_arg_tmp_buffer)), aligned_arg)
```
vchuravy: Hm can't you do:
```
tmp_buffer = Ref{T_aligned}(aligned_arg).
api.clSetKernelArg(k.id, cl_uint(idx - 1), cl_sizeof(T_aligned), tmp_buffer)
```

and in due time this will happen on the stack?

SimonDanisch: Yeah, maybe.. I at least should do some more benchmarks, before doing such a premature optimization...

src/kernel.jl
SimonDanisch: I feel like we're far away from multi threaded kernel launches  ;) but if it happens we can still change it!

src/kernel.jl
SimonDanisch: It's actually almost as fast, but I'm currently benchmarking hot loops with OpenCL kernel calls, and there the allocation actually makes a difference.

yuyichao: The allocation of the Ref will be optimized away soon so please don't use a global buffer.

SimonDanisch: Well, I've been hearing that for more than a year... I can remove the global buffer when the optimization is merged into base and released!

yuyichao: Also, using a global buffer is a premature optimization, using a local allocation is the correct way.

SimonDanisch: Well, I need to check more thoroughly, but time spend in the gc goes down quite a bit for some code I'm testing - and so the execution time. So it doesn't immediately look premature.

yuyichao: > I've been hearing that for more than a year

Who said that? I don't think there was ever an PR ready enough to do that soon previously.

yuyichao: > So it doesn't immediately look premature.

Premature doesn't mean it's not an optimization, just that it's wrong. It's solving a problem in a known bad way where there's a Base PR ready to properly fix it.

SimonDanisch: So why is there still `&` on julia master, if it's known that it's an premature optimization with 0.7 coming up? But, jeez, whatever... I can just copy over that code to GPUArrays...

yuyichao: Because the syntax was added long before stack allocation of `Ref` is brought up and before `RefValue`/`Ref` is even a type. It was never added as an optimization so it was never a premature optimization by definition. Also, without knowing Base is actually ready for stack allocation of `Ref` as well as non-experimental multithreading for more and more cases this won't be a premature optimization either. The correct way to do thing changes so now it is definitely a premature optimization.

SimonDanisch: It still doesn't make sense to rely on non existing optimizations - OpenCL doesn't even run on master.
Anyways, it makes sense that I got the benchmarks wrong - if not I can still add the constant buffer to GPUArrays -  which is the package that needs the benchmarks to look good.

SimonDanisch: okay benchmarks were flawed anyways :)

SimonDanisch:
Okay, merge?

src/kernel.jl
```diff
@@ -84,14 +84,181 @@ function set_arg!(k::Kernel, idx::Integer, arg::LocalMem)
...
+load functions can read a vector from an address aligned to the element type of the vector. The
+vector store functions can write a vector to an address aligned to the element type of the vector.
+"""
+function cl_alignement(x)
+    is_cl_inbuild(x) ? inbuild_alignement(x) : cl_sizeof(x)
+end
+function inbuild_alignement{T}(::Type{T})
+    x = sizeof(T)
+    x == 0 && return 0
+    nextnum = ceil(Int, log2(x))
```
vchuravy: `ispow2(x) ? x : nextpow2(x)`

src/context.jl
```diff
@@ -1,24 +1,60 @@
 # OpenCL.Context
 
+const _ctx_id_seen = Dict{CL_context, Bool}()
```
SimonDanisch: What exactly do you mean by a device reset?
But yeah, I think what we actually need is counting also the julia references.. Will update the PR!

src/context.jl
```diff
@@ -1,24 +1,77 @@
...
+        error("Double free of context id: ", ctx_id)
+    elseif refcount == 1
+        delete!(_ctx_reference_count, ctx_id)
+        return
+    end
+    _ctx_reference_count[ctx_id] = refcount - 1
+    return
+end
+
 type Context <: CLObject
```
vchuravy: The other option would be to turn Context into singelton objects... similar to how we handle `CuContext` in the CUDA packages since reference counting for Julia GC managed objects is possible but might be inaccurate if the finalizers are delayed.

src/kernel.jl
```diff
@@ -84,14 +84,179 @@ function set_arg!(k::Kernel, idx::Integer, arg::LocalMem)
...
+    :($size)
+end
+
+"""
+Converts a Julia type to a correctly padded OpenCL type.
+Turns into a no-op when no convert needed
+"""
+@generated function aligned_convert(x)
+    offset = Ref(0); elements = []; fields = []
+    _aligned_convert!(x, offset, elements, fields, :x)
```
vchuravy: Couldn't `_aligned_convert` simply return `offset, elements, fields` instead of having them passed and mutated?

src/kernel.jl
```diff
@@ -84,14 +84,179 @@ function set_arg!(k::Kernel, idx::Integer, arg::LocalMem)
...
 function set_arg!{T}(k::Kernel, idx::Integer, arg::T)
     @assert idx > 0 "Kernel idx must be bigger 0"
     if !isbits(T) # TODO add more thorough mem layout checks and the clang stuff
         error("Only isbits types allowed. Found: $T")
     end
-    boxed_arg = Ref{T}(arg)
-    @check api.clSetKernelArg(k.id, cl_uint(idx - 1), sizeof(T), boxed_arg)
+    aligned_arg, off = aligned_convert(arg)
+    T_aligned = typeof(aligned_arg)
+    # is this save?
```
vchuravy: This comment shouldn't be necessary anymore.

test/test_context.jl
```diff
@@ -5,7 +5,21 @@
...
+                @test ctx.id != C_NULL
+                @test ctx2.id != C_NULL
                 finalize(ctx)
+                @test ctx.id == C_NULL
+                @test ctx2.id != C_NULL
+                @test cl.is_ctx_id_alive(ctx_id)
+                finalize(ctx2)
+                @test ctx.id == C_NULL
+                @test ctx2.id == C_NULL
+                # jeez, this segfaults... WHY? I suspect a driver bug for refcount == 0?
```
vchuravy: On which drivers did you run into this?

src/context.jl
SimonDanisch: Well, but this is reference counting for correctness and not performance, so it should be fine!
It's just tracking the creation of the finalizer, so acting whenever the finalizer get's called is the right time!
Singleton doesn't sound too bad! We just need to figure out, how this works with different devices and if there are use cases, when people want multiple contexts on the same device...

src/kernel.jl
SimonDanisch: yes, that's just an artifact from my first implementation... I actually just still found a bug in here, so I need to rewrite this anyways -.-

SimonDanisch:
Jeez, this was annoying, especially when considering how simple the rule in the ends are.
But this works now, and also means we can more easily pass a broad range of structs to GPUArrays :)
I hope we can finally merge this :)

test/test_context.jl
SimonDanisch: NVIDIA 381.22

SimonDanisch:
Okay cleaned up the commits! Let's hope all of the commits pass and we can finally merge this :D

PR #142, opened 2017-09-16
Fix depwarns on 0.7

yuyichao:
Deprecate the `@min_v??` macros and replace them with `min_v??` functions since the macro relies
on deprecated syntax on 0.7 (`?` as identifiers).
The macro does not do anything that function can't do and
a function actually makes it much more clear what the code is doing.
Also include `min_v21` and `min_v22` now that the spec is released.


vchuravy:
Thanks! It seems our travis environment is broken again...

PR #144, opened 2017-09-20
support pointefree mutables

SimonDanisch:


src/kernel.jl
```diff
@@ -247,7 +247,7 @@ function set_arg!{T}(k::Kernel, idx::Integer, arg::T)
     ref = if isbits(T_aligned)
         Base.RefValue(packed)
     else
-        packed
+        pointer_from_objref(packed)
```
vchuravy: Wouldn't it be fine to just say: `Ref(packed)`? and then let the logic for references take over instead of re-implementing it here?

SimonDanisch: Yeah, that might just work...

vchuravy: Do you want to quickly fix it otherwise I will merge this and tag a point release.

test/test_behaviour.jl
```diff
@@ -272,4 +272,56 @@ let test_struct = "
         @test all(x -> x == 13.5, r)
     end
 end
+
+end
+
+mutable struct MutableParams
```
vchuravy: This won't work on 0.5. See test failure on GPU CI...

PR #146, opened 2017-10-19
only replace NTuple{3, T} and empty

SimonDanisch:


SimonDanisch:
After finding out that Julia includes padding in structs and that you can actually align fields in OpenCL, 
it turns out one only needs to replace empty types and `NTuple{3, <: Numbers}` in structs and when the alignment differs, you need to include that in the definition of the struct.
With this, much less conversions are needed!

SimonDanisch:
This seems to make inference hang on 0.5 - I played around with it a bit and couldn't get it not to hang without a major effort. Is it time to drop 0.5?

jpsamaroo:
My 2 cents - drop 0.5 support. Julia 0.7/1.0 seems to be coming very soon anyway, so you'll just end up needing to either support 3 major versions at once, or drop the oldest (0.5). And I truly hope that people who use GPUs are using 0.6 or greater to get those performance improvements.

SimonDanisch:
Yeah, at this point it would be kinda self destructive for anyone to stick with 0.5 ;)

SimonDanisch:
aaaaalright, intel has `__attribute__` packed implemented differently from nvidia...

SimonDanisch:
Might as well be a bug: https://software.intel.com/en-us/forums/opencl/topic/280668

SimonDanisch:
lol, turns out the keyword is officially `aligned` and not `align`. Not sure if I should be angry at nvidia for allowing the alternative spelling, at intel for not allowing the same alternative spellings, or at me for misspelling it in the first place....

SimonDanisch:
Okay, since this is blocking a release of CLArrays, I decided to move this new bug out of this pr into #148, so that we can merge this. Hope this is fine with anyone!

SimonDanisch:
Ah damn, I wanted to squash!

PR #153, opened 2018-04-17
try to print error better

SimonDanisch:
One problem this fixes is, that the callback also get's called for warnings. 
In the previous version, this would generate a fatal error.
One problem it's introducing: I disregard user callbacks ;)
Do we want to fix that problem, or is no one using that feature anyways?
We probably should add a deprecation in that case.

### DIFF
--- a/src/context.jl
+++ b/src/context.jl
@@ -89,16 +89,23 @@ immutable _CtxErr
     cb :: Csize_t
 end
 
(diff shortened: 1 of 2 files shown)


SimonDanisch:
I just implemented the custom callback...
Seems like nvidia doesn't really care about those callbacks though: https://devtalk.nvidia.com/default/topic/497433/context-callback-never-called/

So after wasting quite a bit of time to figure out how to test this - I conclude we cannot actually test this -.- bamboozled again!

PR #174, opened 2019-02-18
Test on AMDGPU machine as well

vchuravy:
Thanks to @jpsamaroo we now have a CI machine with an AMD GPU.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -3,66 +3,54 @@ stages:
     - post
 
 include:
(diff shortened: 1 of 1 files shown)


vchuravy:
bors try

jpsamaroo:
@vchuravy given that we do pretty well on most of the Gitlab tests (even if not passing), maybe we should merge this now to assist with testing other PRs (since CI is currently rather sparse)? We could then address recurring failures in a future PR.

PR #177, opened 2019-12-21
Always allow package to load

jpsamaroo:


jpsamaroo:
Amazingly, this passes tests on the NV systems which was not previously the case (of course, I don't think that's because of anything I did here). Regardless, the package appears to load correctly, which is good.

jpsamaroo:
Closes #186

PR #184, opened 2020-03-04
Allow 64-bit buffer lengths

### LINKED ISSUE #183: Cannot create a buffer with a size than exceeds a UInt32
I am unable to create a buffer with a size that exceeds the size of a `UInt32` due to an `InexactError` in `buffer.jl:126` (it currently casts the number of bytes of the buffer to a `cl_uint`). I should be able to allocate a buffer larger than this on my card as `device[:address_bits]` is 64. 

I don't think there would be an issue changing this parameter to type `cl_ulong`, but I have not gone through the code in great detail.

herzfeldd:
Fixes #183

### DIFF
--- a/src/buffer.jl
+++ b/src/buffer.jl
@@ -123,7 +123,7 @@ function Buffer(::Type{T}, ctx::Context, flags::CL_mem_flags,
     end
 
     err_code = Ref{CL_int}()
-    mem_id = api.clCreateBuffer(ctx.id, flags, cl_uint(nbytes),
+    mem_id = api.clCreateBuffer(ctx.id, flags, nbytes,
                                 hostbuf !== nothing ? hostbuf : C_NULL,
                                 err_code)
     if err_code[] != CL_SUCCESS


SimonDanisch:
Seems like we need a `size_t` for this?
```
The unsigned integer type of the result of the sizeof operator. This is a 32-bit unsigned integer if CL_DEVICE_ADDRESS_BITS defined in clGetDeviceInfo is 32-bits and is a 64-bit unsigned integer if CL_DEVICE_ADDRESS_BITS is 64-bits.
```

https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/clCreateBuffer.html

DO we already have this in OpenCL.jl?

herzfeldd:
The underlying call to the C functions already use a `Csize_t` at src/api/opencl_1.0.0.jl#43. 

The cast to either the original `cl_uint` or `cl_long` is not strictly necessary as Julia's `ccall` should do the cast to a `Csize_t` (I have verified that this is true). 

However, unlike the spec that you quoted, `Csize_t` in the `ccall` does not depend on the value of `:address_bits`, but I am not sure this is necessary.

I am happy to replace this pull request with one that merely removes the reference to the original `cl_uint` or `cl_ulong`.

SimonDanisch:
Whatever you think is safe ;) Just trying to make sure we're not overlooking anything!

herzfeldd:
To be consistent with the rest of the code in `buffer.jl`, I removed the explicit cast in its entirety. 

The remainder of the functions (e.g., `enqueue_copy_buffer`, `enqueue_map_mem`) use the result `sizeof(buffer)` and then allow the underlying api to cast the result as a `Csize_t`.

ericproffitt:
I'm not sure where everyone has landed on this, but loading an array into buffer memory with a size that exceeds UInt32 bytes does not appear to work. Technically it can be done with the suggested modification `cl_ulong(nbytes)`, however reading the data back onto the host will show that much of the data has been zeroed out.

juliohm:
Closing as too old

PR #189, opened 2020-12-11
Added quick reference to README.md

mxkramer:
Hello! I noticed the package doesn't have much in the way of documentation outside of examples, and while the API is generally clear from using `methods`, I found it difficult to get started due to some name changes. I thought I'd make a quick reference to the API, generally following the Khronos OpenCL v1.2 quick reference card, with the eventual idea of using `Documenter.jl`. I'm relatively new to OpenCL and especially this package, so please let me know if there are mistakes or improvements I could make.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,3 +103,61 @@ else
     @error "Norm should be 0.0f"
 end
 ```
+
+## Translation
+
+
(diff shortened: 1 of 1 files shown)


jpsamaroo:
Thanks!

Issue #190, opened 2020-12-12
Expose function interface for `cl.info`


miakramer:
I've been playing around with this package, and I think the API might be improved by exposing certain info calls. Internally, many of them use a `Dict` from `Symbol` to `Function`: I'm just proposing exporting the functions in addition. So, you could either call `cl.info(device, :name)` or `cl.name(device)`. This increases the number exported functions by quite a bit, but IMO it's a more ergonomic interface and is type-stable, which is nice for user code which depends on this info (avoids having to typeassert). `cl.info(device, :max_work_group_size)` has return type `Any`, but exporting `cl.max_work_group_size(device)` gives a stable return type of `UInt64`.

I'm happy to do the work (already did `device.jl` just messing around), just wanted to check if this would be a desired change.

jpsamaroo:
That sounds fine to me!

PR #191, opened 2020-12-16
Separated object info APIs

miakramer:
See #190. This commit separates the API for getting info about an object. For example, `cl.info(device, :name)` is also exposed as `name(device)`. This is a more ergonomic API, both in usage and because the return type is (nearly always) stable.

There are two changes to the API, which I can remove if they're unwanted:

- The API of `binaries(program)` changed from `Dict{Device, Array{UInt8, N} where N}` to `Dict{Device, Vector{UInt8}}`, as it seemed a vector would always be returned.
- The API of `work_group_compile_size(device, kernel)` differs from `work_group_info(kernel, :compile_size, device)`: it returns a tuple with known size rather than a vector.

Additionally, each of these new methods has been given at least basic documentation, with the signature and a brief description including the OpenCL info constant name. Compatibility and deprecation info was added in the Documenter.jl style, so eventually it can be used. For now, they look nice in the REPL help.


miakramer:
Oh, weird, I didn't expect anything to fail! It looks like the testing device failed to make a queue for the device. I can test on a Linux machine in a few minutes.

miakramer:
Yeah, so it runs fine on Linux (minus the test that required 64-bit support, the laptop only has intel integrated graphics)

jpsamaroo:
The tests for this package are seriously unreliable, I wouldn't worry about random failures so long as it works locally. I'll review this the next chance I get.

PR #192, opened 2021-08-24
README overhaul

carstenbauer:
I refactored the README.md, most sections of which were written > 5 years ago. While "cleaning up", I dropped the "supported Julia versions" section as it was completely out of date and highly misleading. If we really want such a section, someone should sit down and update it appropriately.

I plan to merge this soon, but please take a look and suggest improvements!

@vchuravy @jpsamaroo

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 # OpenCL.jl
 
-[![][buildbot-julia05-img]][buildbot-julia05-url] [![][buildbot-julia06-img]][buildbot-julia06-url] [![][buildbot-travis-img]][buildbot-travis-url] [![](https://img.shields.io/coveralls/JuliaGPU/OpenCL.jl.png)](https://coveralls.io/r/JuliaGPU/OpenCL.jl?branch=master)
+![][lilfecycle-img]
+[![][buildbot-julia05-img]][buildbot-julia05-url]
+[![][buildbot-julia06-img]][buildbot-julia06-url]
+[![][buildbot-travis-img]][buildbot-travis-url]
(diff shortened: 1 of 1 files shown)


carstenbauer:
One more idea: we could add a Lifecycle badge, or maybe Lifecycle.

jpsamaroo:
It's not really experimental; more like maintenance mode?

README.md
```diff
@@ -13,60 +9,24 @@ OpenCL bindings for Julia
...
-
-```julia
-  Pkg.add("OpenCL")
-```
-
-3. OpenCL will be installed in your ``.julia`` directory
-4. ``cd`` into your ``.julia`` directory to run the tests and try out the examples
-5. To update to the latest development version, from the Julia repl:
+## Installation
+1. Install an OpenCL driver. (If you're on macOS, OpenCL should already be available.)
```
jpsamaroo, line 19: We should specify what versions of macOS are supported (since OpenCL support went away in macOS recently).

carstenbauer:
> It's not really experimental; more like maintenance mode?

Yeah, I agree. But is there a badge for this?

README.md
carstenbauer: Hm, I'm on macOS Big Sur and and it works, but maybe I installed it with some other software. Anyways, we could just drop the macOS comment entirely to be safe. What do you think?

jpsamaroo: We should probably say instead, "If you're on macOS, OpenCL is either already available or unsupported"?

carstenbauer:
Incorporated the comments. Feel free to make further modifications in seperate commits / PRs.

PR #194, opened 2022-01-22
use VersionParsing

stevengj:
I noticed that you were doing your own version-number parsing, and thought you might want to use the VersionParsing parsing package instead for more flexibility.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -6,6 +6,7 @@ version = "0.8.1"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
+VersionParsing = "81def892-9a0e-5fdd-b105-ffc91e053289"
 
 [extras]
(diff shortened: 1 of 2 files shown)


juliohm:
@stevengj we are now using OpenCL_jll as a dependency to download the library on different platforms for end-users. We are also planning to use Clang.jl to create the API file, does it make sense? Can you comment on the API lookup that is implemented here? I am trying to revive the package and am not familiar with what was done here in the past.

stevengj:
The jll package should already have the version number as a `VersionNumber`, I think?

juliohm:
It has a version but I'm not sure if this version is the version of the
OpenCL spec discussed in this PR? Basically OpenCL_jll has the ICD loader,
which as far as I understand supports all OpenCL specs. The OpenCL driver
implementation is the one that contains the version maybe? I'm considering
PoCL as a generic driver to install as an artifact. That way whenever
people install OpenCL.jll they will get the loader OpenCL_jll and a
portable PoCL driver.

Em qua., 5 de out. de 2022 21:08, Steven G. Johnson <
***@***.***> escreveu:

juliohm:
I will close the PR assuming that we will handle versions in a more "automatic" way moving forward.

Thank you for replying. Hope we can revive OpenCL.jl soon 🙏🏽

PR #196, opened 2022-10-04
Add OpenCL_jll dependence

ClaroHenrique:
This PR aims to add the OpenCL_jll dependency.
With that, the host machine dont need to have OpenCL installed.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,8 +3,8 @@ uuid = "08131aa3-fb12-5dee-8b74-c09406e224a2"
 version = "0.8.1"
 
 [deps]
-Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+OpenCL_jll = "6cb37087-e8b6-5417-8430-1f242f1e46e4"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 
 [extras]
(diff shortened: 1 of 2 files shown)


Project.toml
```diff
@@ -3,8 +3,8 @@ uuid = "08131aa3-fb12-5dee-8b74-c09406e224a2"
 version = "0.8.1"
 
 [deps]
-Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+OpenCL_jll = "6cb37087-e8b6-5417-8430-1f242f1e46e4"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
```
juliohm, line 9: I think we also need to add a [compat] section

src/api.jl
```diff
@@ -2,11 +2,9 @@ module api
 
 include("types.jl")
 
-const paths = Sys.isapple() ? String["/System/Library/Frameworks/OpenCL.framework"] : String[]
+import OpenCL_jll
```
juliohm, line 5: It is great to see this hard-coded string gone

Project.toml
ClaroHenrique: would be like this ?
```
[compat]
OpenCL_jll = "2022.9.23"
```

juliohm: Yes, I think that is right. Can you update the PR so that I can merge it?

juliohm:
I am getting the following error with this PR:

```julia
julia> using OpenCL
julia> cl.platforms()
ERROR: CLError(code=-1001, CL_PLATFORM_NOT_FOUND_KHR)
```

juliohm:
As we discussed on our call, this error is happening because of the missing OpenCL driver. Everything is working after a driver is manually installed.

PR #197, opened 2022-10-04
Add Buildkite CI

ClaroHenrique:
This PR aims to add Buildkite CI support and remove the deprecated Gitlab CI.


juliohm:
@ClaroHenrique can you comment on how close we are to get a final PR here? Can we merge this to see the build happening on buildkite?

ClaroHenrique:
> @ClaroHenrique can you comment on how close we are to get a final PR here? Can we merge this to see the build happening on buildkite?

The Buildkite pipeline was recently created: https://buildkite.com/julialang/opencl-dot-jl
So we can start developing the tests.

juliohm:
We need to install an OpenCL driver in the buildkite image 👍🏽

test/test_platform.jl
```diff
@@ -2,14 +2,17 @@
...
+            @testset "Platform $(p[:profile])" begin
+                @test p != nothing
+                @test pointer(p) != C_NULL
+                for k in [:profile, :version, :name, :vendor, :extensions]
+                    @test p[k] == cl.info(p, k)
+                end
+                # ignore version tests until OpenCL 3 support is available
+                #v = cl.opencl_version(p)
+                #@test 1 <= v.major <= 2
+                #@test 0 <= v.minor <= 2
```
juliohm, line 14: Is there a reason why this code block is indented differently? It seems that it only has whitespace changes?

ClaroHenrique: I added the `@testset "Platform $(p[:profile])" begin` block to get a better log when a test fail.

juliohm: Yes, but I think the indentation was changed accidentally? The diff shows various lines of code that are identical, just shifted to the right. Can you please adjust that?

juliohm: Also, I think commenting the test on opencl_version isn't ideal. You can just adapt it and replace the upper bound by 3?

test/test_program.jl
```diff
@@ -43,23 +43,25 @@
...
-                @warn("Skipping OpenCL.Program build for Portable Computing Language Platform")
-                continue
-            end
-            @test prg[:build_status][device] == cl.CL_BUILD_SUCCESS
+                # BUILD_SUCCESS undefined in POCL implementation..
+                if device[:platform][:name] == "Portable Computing Language"
+                    @warn("Skipping OpenCL.Program build for Portable Computing Language Platform")
+                    continue
+                end
+                @test prg[:build_status][device] == cl.CL_BUILD_SUCCESS
```
juliohm, line 56: Can you please double check here if you introduced TABS acccidentally?

test/test_platform.jl
juliohm: Oh nevermind about the indentation. I just saw that you introduced a new testset.

juliohm (changes requested):
We just need to double check the commented test and the TABs.

test/test_platform.jl
ClaroHenrique: > Also, I think commenting the test on opencl_version isn't ideal. You can just adapt it and replace the upper bound by 3?

done

test/test_program.jl
ClaroHenrique: Its ok, this code was also inserted inside a testset block.

juliohm (approved):

PR #264, opened 2024-10-25
Revamp memory management, and add USM support.

VarLad:
~~**Heavy WIP, for now. Expect very little to work.**~~

~~This exists so that I can, with time, ask for feedbacks and at a later point, for other people to be able to try this PR.~~

Edit:

Most of the functionality is there now. ~~There's a bug in array interface, where operations between two `CLArray` results in an error.~~ This has been fixed in the latest commit.

~~Most recent update: It should be on par with former SVM backend, though tests haven't been ported yet. Needs testing!~~

~~Most recent update: 79 errors in GPUArrays test suite. All errors basically say: `Attempt to use a freed reference`~~

~~All tests pass! Should be usable as of the moment!~~

~~Most recent update: Going into WIP again~~

Most recent update: All tests pass as per latest commit. Feel free to try!


VarLad:
Tests still need to be ported, but any reviews in current functionality would be welcome!

lib/cl/CL.jl
```diff
@@ -15,9 +17,9 @@ include("device.jl")
 include("context.jl")
 include("cmdqueue.jl")
 include("event.jl")
+include("pointer.jl")
+include("intelfns.jl")
 include("memory.jl")
-include("buffer.jl")
-include("svm.jl")
```
maleadt, line 20: Why did you remove this functionality?

lib/cl/api.jl
```diff
@@ -66,7 +66,7 @@ function retry_reclaim(f, isfailed)
         end
     end
 
-    ret
+    return ret
```
maleadt, line 69: Unrelated formatting changes like that make it hard to review your PR. The formatter should only have run on changed code; did you format the entire codebase?

lib/cl/api.jl
```diff
@@ -148,7 +148,7 @@ const initialized = Ref{Bool}(false)
         if is_high_integrity_level()
             @warn """Running at high integrity level, preventing OpenCL.jl from loading drivers from JLLs.
 
-                     Only system drivers will be available. To enable JLL drivers, do not run Julia as an administrator."""
+            Only system drivers will be available. To enable JLL drivers, do not run Julia as an administrator."""
```
maleadt, line 151: To prevent the formatter from spamming, keep your PR as draft for longer.

lib/cl/state.jl
```diff
@@ -139,9 +140,121 @@ end
...
+end
+
+
+@inline function active_state()
+    # inline to remove unused state properties
+    state = task_local_state!()
+    return (device=state.device, context=state.context, queue=queue(state),
+            math_mode=state.math_mode, math_precision=state.math_precision)
+end
+=#
```
maleadt, line 260: Remove this.

lib/cl/cmdqueue.jl
```diff
@@ -81,12 +85,23 @@ function Base.getproperty(q::CmdQueue, s::Symbol)
...
     end
 end
 
+function global_queue(ctx::Context, dev::Device)
+    # NOTE: dev purposefully does not default to context() or device() to stress that
+    #       objects should track ownership, and not rely on implicit global state.
+    return get!(task_local_storage(), (:CLCommandQueue, ctx, dev)) do
+        CmdQueue(ctx, dev)
+    end
+end
```
maleadt, line 104: What's wrong with `cl.queue` from `state.jl`?

lib/cl/error.jl
```diff
@@ -87,80 +87,92 @@ const _cl_error_codes = Dict{Int, Symbol}(
...
 
     CL_INVALID_HOST_PTR =>
-    string("If host_ptr is NULL and CL_MEM_USE_HOST_PTR or ",
-           "CL_MEM_COPY_HOST_PTR are set in flags or if host_ptr is not NULL but ",
-           "CL_MEM_COPY_HOST_PTR or CL_MEM_USE_HOST_PTR are not set in flags."),
+        string(
+        "If host_ptr is NULL and CL_MEM_USE_HOST_PTR or ",
+        "CL_MEM_COPY_HOST_PTR are set in flags or if host_ptr is not NULL but ",
+        "CL_MEM_COPY_HOST_PTR or CL_MEM_USE_HOST_PTR are not set in flags."
+    ),
```
maleadt, line 103: Undo all the formatting changes here.

lib/cl/kernel.jl
```diff
@@ -51,60 +51,64 @@ Base.length(l::LocalMem{T}) where {T} = Int(l.nbytes ÷ sizeof(T))
...
 end
 
 # SVMBuffers
 ## when passing using `cl.call`
-function set_arg!(k::Kernel, idx::Integer, arg::SVMBuffer)
-    clSetKernelArgSVMPointer(k, cl_uint(idx-1), arg.ptr)
+function set_arg!(k::Kernel, idx::Integer, arg::AbstractBuffer)
+    ext_clSetKernelArgMemPointerINTEL(k, cl_uint(idx - 1), arg.ptr)
     return k
 end
```
maleadt, line 67: This isn't SVM anymore.

maleadt (changes requested):
Quick partial review. My biggest gripes are that this shouldn't have formatted the entire codebase, which makes reviewing very hard, and shouldn't have removed support for regular Buffers or SVM. Intel USM being an extension I don't think we should rely on it being available (SVM is questionable enough already, but at least is mandated by specific OpenCL versions).

lib/cl/api.jl
VarLad: I tried running Runic on the files it complained.

lib/cl/cmdqueue.jl
VarLad: This is a testing artifact, we don't use or need this, will remove this.

maleadt:
```
julia> cl.svm_capabilities(cl.device())
(coarse_grain_buffer = true, fine_grain_buffer = false, fine_grain_system = false)

julia> cl.usm_capabilities(cl.device())
ERROR: CLError(code=-30, CL_INVALID_VALUE)
```

That should probably behave more nicely when USM isn't supported at all.

maleadt (changes requested):

maleadt:
Something is wrong with CI (on master too); the CUDA tests are empty.

maleadt:
I've squashed the commits. Please use rebase in the future; merge commits only make things more annoying.

lib/cl/memory.jl
```diff
@@ -1,3 +1,315 @@
+# Raw memory management
+
+export device_alloc, host_alloc, shared_alloc, free #, properties, lookup_alloc
```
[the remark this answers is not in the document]
maleadt: Feel free to mark conversations as resolved when you fixed something, that makes it easier to review.

maleadt:
The Memoization.jl dependency isn't great, also because it doesn't offer us thread safety (unless when using a ridiculously expensive datatype, like ThreadSafeDicts.jl). I wonder if we should copy the CUDA.jl LazyInitialized/`@memoize` implementation. Alternatively, for now we could just put the buffer type in the task local state, initialized when switching devices.

I'll have a closer look tomorrow.

maleadt:
I'm seeing lots of failures locally when testing non-PoCL platforms:

```
      From worker 2:	error in running finalizer: UndefRefError()
      From worker 2:	ext_clMemBlockingFreeINTEL at /home/tim/Julia/pkg/OpenCL/lib/cl/modified_fns.jl:34
      From worker 2:	#free#401 at /home/tim/Julia/pkg/OpenCL/lib/cl/memory/memory.jl:43 [inlined]
      From worker 2:	free at /home/tim/Julia/pkg/OpenCL/lib/cl/memory/memory.jl:28 [inlined]
      From worker 2:	release at /home/tim/Julia/pkg/OpenCL/src/pool.jl:78 [inlined]
      From worker 2:	#9 at /home/tim/Julia/pkg/OpenCL/src/array.jl:55
      From worker 2:	unknown function (ip: 0x71b84b70a752)
      From worker 2:	release at /home/tim/.julia/packages/GPUArrays/Mot2g/src/host/abstractarray.jl:42
      From worker 2:	unsafe_free! at /home/tim/.julia/packages/GPUArrays/Mot2g/src/host/abstractarray.jl:100 [inlined]
      From worker 2:	unsafe_free! at /home/tim/.julia/packages/GPUArrays/Mot2g/src/host/abstractarray.jl:115
      From worker 2:	unknown function (ip: 0x71b84b70a372)
      From worker 2:	run_finalizer at /cache/build/tester-amdci5-12/julialang/julia-release-1-dot-11/src/gc.c:299
```

I've squashed again and reverted all the unnecessary formatting changes. Will be taking another look later today / this week.

VarLad:
@maleadt Can you try running those tests individually? As in, `cd test; include("setup.jl"); include("kernel.jl")`
and see if you still encounter the same errors?
or alternatively, try removing `pocl_jll` from `using OpenCL, pocl_jll` in `tests/setup.jl`.
I have a feeling that the error is because of different processes are choosing context/devices differently/randomly...

VarLad:
I tried printing some logs and I see this:
image

and I can't really make sense of this. Both are printing `cl.device()`
One in `test/setup.jl`, other in `test/array.jl`

Also, when I remove `pocl_jll` from testing, or test individually, it works out fine.

VarLad:
Ah, thanks for the tips. 
That is more convenient. I see that in runtests you cycle through different platforms and device...
I think thats where the error is coming from. 

For a solution I guess the functions defining CLArray need to be rerun everytime device changes?

maleadt:
The problem is probably that you're calling the wrong extension functions, since you simply look up `cl.platform()` in `ocl_extension`. The `free` function would probably have to switch to the correct platform as derived from the context in the buffer.

lib/cl/memory/svm.jl
VarLad, line 1: Apologies for annoying you on this....
Regarding this change on line 35 (or 36)
```julia
function svm_free(buf::SharedVirtualMemory)
    clSVMFree(context(buf), buf)
    return
end
```
change.
Is that fine? As per https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clSVMFree.html
> Note that clSVMFree does not wait for previously enqueued commands that may be using svm_pointer to finish before freeing svm_pointer. It is the responsibility of the application to make sure that enqueued commands that use svm_pointer have finished before freeing svm_pointer. This can be done by enqueuing a blocking operation such as clFinish, clWaitForEvents, clEnqueueReadBuffer or by registering a callback with the events associated with enqueued commands and when the last enqueued command has finished freeing svm_pointer.

Shouldn't we try and keep the behavior close to the USM counterpart for the `free` function?

VarLad: Ah wait, never mind, you do the waiting part later down I guess.

maleadt: No actually that's a good comment; we need synchronize _before_ calling the free AFAIU.

maleadt:
Doesn't work on Intel, ironically:

```
❯ jl --project -e 'using OpenCL; cl.platform!("intel"); CLArray([1]) .+ 1'
Precompiling OpenCL...
  1 dependency successfully precompiled in 3 seconds. 64 already precompiled.
ERROR: CLError(code=-30, CL_INVALID_VALUE)
Stacktrace:
  [1] throw_api_error(res::Int32)
    @ OpenCL.cl ~/Julia/pkg/OpenCL/lib/cl/libopencl.jl:6
  [2] check
    @ ~/Julia/pkg/OpenCL/lib/cl/libopencl.jl:17 [inlined]
  [3] clSetKernelExecInfo
    @ ~/Julia/pkg/OpenCL/lib/cl/api.jl:34 [inlined]
```

maleadt:
No; turns out we weren't properly pushing to `svm_pointers` on master, so it didn't happen to call `clSetKernelExecInfo`.

maleadt:
This is the failing code path: https://github.com/intel/compute-runtime/blob/e48b52814958331d46bc6fb6398d6793fdc65049/opencl/source/api/api.cpp#L5300-L5302

Somehow the allocated pointer isn't recognized by the driver.

VarLad:
and this doesn't work for USM either?
Can we open a bug report on their side for this? It works with POCL and heck even with NVIDIA drivers...
Also, is your `platform` an Intel CPU or an Intel GPU? I think their CPU drivers might be problematic...
I'll try getting my hand on an Intel iGPU.

maleadt:
Alright, I think I finally got this in a place I'm happy with. It took a lot more work to get all the pieces (calling OpenCL extension functions, introduction of CLPtr, `Managed` and auto-synchronization, USM on PoCL as well as Intel, SVM and USM co-existence, etc) fully working and in a maintainable shape.

@VarLad Thanks for laying the groundwork here! For future work, I would suggest however to keep the scope of PRs minimal and create multiple ones. That will make it much easier to review as well as get everything working.

Feel free to give this a round of review before merging.

VarLad:
Does a version bump make sense for this PR? :slightly_smiling_face:

maleadt:
A release for sure, but in the pre-1.0 regime we can't do a minor release (which is considered breaking). So it'll be a patch bump.

PR #292, opened 2025-03-21
Typo fix : pivate -> private

blegat:
The typo is actually coming from Hands on OpenCL but we might as well just fix it here


maleadt (approved):
Thanks!

Issue #281, opened 2025-01-24
Support `cl_mem` based buffers using `cl_ext_buffer_device_address` extension


VarLad:
More on that here: https://github.com/KhronosGroup/OpenCL-Docs/pull/1159

PR #311, opened 2025-05-31
Add Buffer Device Address Backend

VarLad:
More on the extension here: https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#cl_ext_buffer_device_address

This PR introduces a new backend for the array interface, as well as bumps the generated wrapper version for OpenCL Headers.

Main purpose for the new backend is to be able to support more devices than currently possible. 
This is heavily inspired by the recent merge of bda into Mesa's Rusticl. Specifically the Zink backend, which means  OpenCL.jl can use Vulkan as a backend and ideally work on a large number of previously unsupported devices.

~~**This is still a work in progress, and mostly meant for review or question purposes at the current stage.**~~

Current status: All tests pass. Feel free to try! This is supported both by POCL 7.0 as well as with mesa-git using the environment variable `RUSTICL_ENABLE=zink` (this can work with any compliant Vulkan driver)


VarLad:
Tested with mesa-git rusticl with `RUSTICL_ENABLE=zink` environment variable on an Intel iGPU.
The results are as follows:
image

Indexing looks bad.

It would be nice if I could somehow test the GPUArrays suite with POCL's BDA backend.

VarLad:
@maleadt Can you take a look at the PR to see whats missing? Most tests pass now. 
A few don't but I need to find out if its Mesa or OpenCL.jl, and a few errors I think are acceptable since I'm testing for BDA and Mesa Zink doesn't support SVM or USM.

Regarding BDA specific tests, I think what we have for normal `Buffer`/`cl_mem` should suffice for BDA as well for now :)

Here's the latest one:

image

VarLad:
closes https://github.com/JuliaGPU/OpenCL.jl/issues/324 and https://github.com/JuliaGPU/OpenCL.jl/issues/281

lib/cl/libopencl.jl
```diff
@@ -1263,424 +1263,6 @@ end
 
 const cl_device_partition_property_ext = cl_ulong
 
-const cl_device_command_buffer_capabilities_khr = cl_bitfield
```
maleadt, line 1266: How come all these other extension methods are gone from the wrappers now?

lib/cl/memory/bda.jl
```diff
@@ -0,0 +1,69 @@
...
+    id::cl_mem
+    ptr::CLPtr{Cvoid}
+    bytesize::Int
+    context::Context
+end
+
+function bda_alloc(bytesize::Integer;
+        alignment::Integer = 0, device_access::Symbol = :rw, host_access::Symbol = :rw
+    )
+    bytesize == 0 && error("size 0 is not supported for BufferDeviceMemory.")
```
maleadt, line 11: Can we return a dummy object instead? Zero-size allocations are generally expected to be supported by the layers above, so if this back-end doesn't support it we should somehow spoof the object (e.g. putting in a null pointer, as we do in CUDA).

src/memory.jl
```diff
@@ -38,9 +38,9 @@ function maybe_synchronize(managed::Managed)
     return nothing
 end
 
-function Base.convert(::Type{CLPtr{T}}, managed::Managed{M}) where {T, M}
+function Base.convert(t::Union{Type{CLPtr{T}}, Type{cl.cl_mem}}, managed::Managed{M}) where {T, M}
```
maleadt, line 41: Why do we need to extend this (and other) methods to support `cl_mem`? Don't we always use the device pointer that's part of the BDA, only using the underlying `cl_mem` to free etc?

maleadt (commented):
Couple of initial thoughts. I haven't had the time to thoroughly test this.

maleadt:
If PoCL supports this, we should probably have a way to pick the back-end (e.g. using a preference, asserting at run time that the device supports said back-end) and test this functionality on CI.

The test failures are problematic though. Indexing really shouldn't behave differently on this back-end.

VarLad:
> Indexing really shouldn't behave differently on this back-end.

I fixed the indexing part. :)

The only remaining errors come from:
- Empty arrays of size 0
- There's a bug involving views. views don't work when range is provided. 
For example, `@view x[3:4]` doesn't work (while `@view x[[3,4]]` does). It doesn't give an error, it always treats it as `@view x[1:2]`, that is, it always starts from the first element for some reason...
Would you know the issue or where to look for it?

VarLad:
> If PoCL supports this, we should probably have a way to pick the back-end (e.g. using a preference, asserting at run time that the device supports said back-end) and test this functionality on CI.

PoCL does support this.

src/memory.jl
maleadt: Oh right, the `memcpy` etc. methods are back-end specific (in CUDA we have simple `memcpy`s that support all kind of memory).

lib/cl/memory/bda.jl
VarLad: > In CUDA.jl I've tried to confine all that logic to the high-level wrappers in src

Apparently we already do that in OpenCL, but it was still letting many edge cases slip. 
I've pushed a solution to this in the latest commit. Do let me know if you see issues with it. 

That being said, kernelabstractions and event test just segfaults on Mesa-git. I don't think they depend on BDA, so its probably a Mesa bug. 
There's 2 errors in base, but I think its because of the mentioned 'clarray views with range index' bug.

lib/cl/memory/bda.jl
```diff
@@ -5,10 +5,12 @@ struct BufferDeviceMemory <: AbstractMemory
     context::Context
 end
 
+BufferDeviceMemory() = BufferDeviceMemory(C_NULL, CL_NULL, 0, context())
+
 function bda_alloc(bytesize::Integer;
         alignment::Integer = 0, device_access::Symbol = :rw, host_access::Symbol = :rw
     )
-    bytesize == 0 && error("size 0 is not supported for BufferDeviceMemory.")
```
maleadt: Can you do this from `alloc` in `src/memory.jl` instead? We probably ought to do this for other memories as well, in order to flush out issues more quickly, and be compatible with hypothetical drivers not supporting zero-size allocations (unless this is mandated by the spec).

VarLad:
@maleadt For now, I've introduced a environment variable based method to select backend before loading OpenCL. This should be enough to run pocl + bda tests. 
A better way to do this (using Preferences.jl for example) would be better in another PR in my opinion. 

**All tests pass on my side**

The PR should be ready if you don't have any issues. :)

maleadt:
Taking a look...

VarLad:
Ah, thanks a lot for the Preferences and CI work :smile:

maleadt (changes requested):
This is starting to look good, thanks for the work here.

I do think that `bda.jl` is unnecessarily duplicating the existing functionality from `buffer.jl`. There's essentially no difference between a buffer that's managed by the BDA memory back-end, and a plain old OpenCL buffer, so I think we should reuse all of that existing support (though adding support to the `Buffer{T}` constructor for mapping optional properties to a call to `clCreateBufferWithProperties`), deriving the memory address only when needed. This should do away with all of the `Union{Buffer, cl_mem}` changes in here, because we'd simply be passing `Buffer`s around.

One potential quirk is that `Buffer` is currently typed, but we could simply use `Buffer{UInt8}` in the array code. Over time, it may be interesting to try and make `Buffer` untyped, but IIRC I tried that already and there were some complications.

VarLad:
@maleadt 
- One other issue is freeing being part of finalizer in Buffer construction. Ideally we'd like to use the infra for `Managed` for all potential backends. This is why bda.jl doesn't directly use `Buffer{T}`.
- And as you mentioned, there's the typed Buffer constructor issue.

Some modification to buffer.jl would be required to solve all these issues, like a separate `free` function to manually free memory.

One thing I can think of, is to keep the `Union{cl_mem, Buffer}` in the `enqueue_` functions as is, and separate the `Buffer` constructor function into two parts, such that one function (called `mem_alloc`) that returns a `cl_mem` and a minimal `Buffer` function which just calls this `mem_alloc` function. `bda_alloc` can also use the same function to do its stuff. Most of the duplication is in `bda_alloc`. For the rest, I don't think there is much unnecessary duplication although I might be wrong. The `enqueue_bda_` wrappers exist specifically for GPUArrays compatibility so I don't think they make much sense to directly have in `Buffer`, and the `bda_free` function is there because `Buffer` has its own (automatic) free mechanism.
As for `clCreateBufferWithProperties`, I think this should be the default function to create a buffer (in the new `mem_alloc` function) anyway as its required by OpenCL 3.0, replacing `clCreateBuffer`.

What do you think? Should I try splitting the `Buffer` constructor function?

VarLad:
One other thing I'm conflicted about is the automatic freeing of `Buffer`.
Wouldn't manual free make more sense? Thats more in line with how it'd be done on other platforms, as well as how we do it with other backends. Just a thought.

maleadt:
Yeah, I was considering that too.

VarLad:
I'm quite surprised that tests are passing, given that there's no mechanism to free buffers...

maleadt:
I have some fixes locally. Am trying to unify some more things, I've always been annoyed how the Buffer abstraction is pretty disjoint from our Memory wrappers.

maleadt:
> _at the very least_, SVM is mandated by the actual OpenCL spec.

I don't think it is, at least not for OpenCL 3.0+: https://github.com/KhronosGroup/OpenCL-Docs/pull/913

VarLad:
> obviously only supporting OpenCL C

I believe this will improve in the future if/when https://github.com/JuliaGPU/OpenCL.jl/pull/319 works?

lib/cl/libopencl.jl
maleadt: Yes, but if I generate them like that I get many more. Let my push my version.

maleadt: Actually, no, after updating the headers JLL I get the same result. Weird.

maleadt: Oh, this is https://github.com/KhronosGroup/OpenCL-Headers/pull/268

VarLad:
@maleadt I think clCreateBufferWithProperties is provided by OpenCL 3.0 so maybe we can use just that? (I think this is the case but its confusing to check...)

Other than this, the code looks good, no obvious issues I can see.

maleadt:
> clCreateBufferWithProperties is provided by OpenCL 3.0 so maybe we can use just that and remove `clCreateBuffer`?

Doesn't hurt to keep the old `clCreateBuffer` path, in case somebody wants to reinstate pre-3.0 support.

VarLad:
Ok, makes sense. Looks good to me, lets merge this! :eyes:

PR #314, opened 2025-06-05
(do not merge) Test pocl

christiangnrd:


christiangnrd:
I think I figured it out. Details elsewhere but the gist is I think these tests only pass on macOS when Command Line Tools are installed.

PR #345, opened 2025-07-24
fix broadcasting for differently backed CLArrays of different dimensions

simeonschaub:


simeonschaub:
Hmm, are there backends where the default `CLArray` constructor does not create one backed by `cl.UnifiedDeviceMemory`?

simeonschaub:
Looks like CI still needs to be approved. Otherwise this should be good to go once tests pass

maleadt (approved):
CI failures look unrelated indeed. Thanks for the PR!

PR #363, opened 2025-09-04
add profiling docs

simeonschaub:


maleadt (approved):
Any way to do interactive profiling?

simeonschaub:
Not AFAIK, doesn't look like perfetto supports it

PR #365, opened 2025-09-06
implement device-side RNG

simeonschaub:
Requires https://github.com/JuliaGPU/GPUCompiler.jl/pull/717

### DIFF
--- a/src/compiler/compilation.jl
+++ b/src/compiler/compilation.jl
@@ -29,7 +29,7 @@ function GPUCompiler.finish_module!(@nospecialize(job::OpenCLCompilerJob),
 
     # if this kernel uses our RNG, we should prime the shared state.
     # XXX: these transformations should really happen at the Julia IR level...
-    if callconv(entry) == LLVM.API.LLVMSPIRKERNELCallConv #haskey(globals(mod), "global_random_keys")
+    if callconv(entry) == LLVM.API.LLVMSPIRKERNELCallConv && haskey(functions(mod), "julia.gpu.additional_arg_getter")
         # insert call to `initialize_rng_state`
         f = initialize_rng_state
         ft = typeof(f)
--- a/src/compiler/compilation.jl
+++ b/src/compiler/compilation.jl
@@ -82,7 +82,6 @@ function GPUCompiler.finish_module!(@nospecialize(job::OpenCLCompilerJob),
 
         # XXX: put some of the above behind GPUCompiler abstractions
         #      (e.g., a compile-time version of `deferred_codegen`)
-
     end
     return entry
 end
--- a/src/compiler/compilation.jl
+++ b/src/compiler/compilation.jl
@@ -18,6 +18,96 @@ GPUCompiler.isintrinsic(job::OpenCLCompilerJob, fn::String) =
     in(fn, known_intrinsics) ||
     contains(fn, "__spirv_")
 
+GPUCompiler.kernel_state_type(job::OpenCLCompilerJob) = KernelState
+
+function GPUCompiler.finish_module!(@nospecialize(job::OpenCLCompilerJob),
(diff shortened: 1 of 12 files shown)


simeonschaub:
The issue is that to know how much memory we are going to need, we need to query the number of subgroups inside a workgroup. I don't think we can do that for local memory declared inside a module.

src/device/runtime.jl
```diff
@@ -12,3 +12,54 @@ report_exception(ex) = return
...
+        end
+
+        call_function(llvm_f, state)
+    end
+end
+
+for name in [:random_keys, :random_counters]
+    @eval @inline @generated $name() =
+        additional_arg_value(LLVMPtr{UInt32, AS.Workgroup}, $(String(name)))
+end
```
maleadt: Can you put these in their own "section" and comment the purpose, for when we revisit this code in the future?

test/Project.toml
```diff
@@ -3,6 +3,7 @@ Adapt = "79e6a3ab-5dfb-504d-930d-738a2a938a0e"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 GPUArrays = "0c68f7d7-f131-5f86-a1c3-88cf8149b2d7"
+GPUCompiler = "61eb1bfa-7361-4325-ad38-22787b887f55"
```
maleadt: I don't see added tests that rely on GPUCompiler?

src/compiler/compilation.jl
```diff
@@ -18,6 +18,83 @@ GPUCompiler.isintrinsic(job::OpenCLCompilerJob, fn::String) =
...
+        #      (e.g., a compile-time version of `deferred_codegen`)
+    end
+    return entry
+end
+
+function GPUCompiler.finish_linked_module!(@nospecialize(job::OpenCLCompilerJob), mod::LLVM.Module)
+    for f in GPUCompiler.kernels(mod)
+        kernel_intrinsics = Dict(
+            "julia.spirv.random_keys" => (; name = "random_keys", typ = LLVMPtr{UInt32, AS.Workgroup}),
+            "julia.spirv.random_counters" => (; name = "random_counters", typ = LLVMPtr{UInt32, AS.Workgroup}),
```
maleadt: These are entirely OpenCL.jl-specific, so maybe name them as such?

```suggestion
            "julia.opencl.random_keys" => (; name = "random_keys", typ = LLVMPtr{UInt32, AS.Workgroup}),
            "julia.opencl.random_counters" => (; name = "random_counters", typ = LLVMPtr{UInt32, AS.Workgroup}),
```

src/compiler/execution.jl
```diff
@@ -151,12 +151,16 @@ abstract type AbstractKernel{F, TT} end
...
+    pushfirst!(call_t, KernelState)
+    pushfirst!(call_args, :(KernelState(Base.rand(UInt32))))
+
     # finalize types
     call_tt = Base.to_tuple_type(call_t)
 
     quote
         indirect_memory = cl.AbstractMemory[]
-        clcall(kernel.fun, $call_tt, $(call_args...); indirect_memory, call_kwargs...)
+        device_rng = kernel.fun.num_args == $(length(call_args) + 2)
```
maleadt: This is ugly. Could we instead store a flag in the HostKernel object that indicates the RNG is in use? `fun.num_args` requires an API call I'd like to avoid in the kernel launch hot path.

maleadt (commented):
> So if we don't calculate this for each kernel launch, we would overallocate around 1 kB of local memory for each work group.

I guess, I just don't like the implementation, hacking in additional arguments in random places. Probably fine for now, but I hope we find something better.

simeonschaub:
I believe this is ready now, ok to merge?

PR #368, opened 2025-09-12
add Float16 math functions

### LINKED ISSUE #374: wrong definition for `rem`
`remainder` is described in https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html as:

> Compute the value r such that r = x - n*y, where n is the integer nearest the exact value of x/y. If there are two integers closest to x/y, n shall be the even one. If r is zero, it is given the same sign as x.

This is different to Julia's `rem` where `n` is `x/y` rounded to zero, so we can't use the `remainder` intrinsic here

simeonschaub:
also fixes #374 by simply removing that device override

### DIFF
--- a/lib/intrinsics/Project.toml
+++ b/lib/intrinsics/Project.toml
@@ -1,7 +1,7 @@
 name = "SPIRVIntrinsics"
 uuid = "71d1d633-e7e8-4a92-83a1-de8814b09ba8"
 authors = ["Tim Besard <tim.besard@gmail.com>"]
-version = "0.5.2"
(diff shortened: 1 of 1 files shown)


simeonschaub:
Ah, I see #369 is failing with the same `copysign` errors - though not the `hypot` ones. Looks like this is just another instance of https://github.com/JuliaGPU/OpenCL.jl/issues/312 then. Since we don't build the jll with Float16 support, we probably shouldn't test the local builds with Float16 either.

PR #369, opened 2025-09-15
allow disabling validation

simeonschaub:
While writing tests for eschnett/SIMD.jl#147, I noticed that `spirv-val` doesn't seem to understand the `SPV_INTEL_masked_gather_scatter` extension and fails if it encounters vectors containing pointers. Ideally should be fixed upstream, but add an option to disable validation for now.


simeonschaub:
Ah, good catch! Looks like the patch is not yet in 2025.1, will open a PR on Yggdrasil

PR #371, opened 2025-09-22
replace copysign.f16 for pocl backend

simeonschaub:
companion PR to JuliaGPU/GPUCompiler.jl#722


pjaaskel:
Does this mean that PoCL misses copysign for FP16 or it's broken somehow? Please file an issue to PoCL if that's the case.

simeonschaub:
> Does this mean that PoCL misses copysign for FP16 or it's broken somehow? Please file an issue to PoCL if that's the case.

Yes, it's missing the intrinsic for it. Opened https://github.com/pocl/pocl/issues/2030

PR #379, opened 2025-10-10
SIMD support for math intrinsics

### LINKED ISSUE #376: support vectorized math intrinsics
It should be fairly straightforward to support those vector intrinsics by adding a SIMD.jl extension to SPIRVIntrinsics

simeonschaub:
closes #376


maleadt (approved):
Interesting!

PR #392, opened 2025-10-15
Use `--quickfail` in CI

christiangnrd:
This should reduce the CI burden for non-hanging tests.


simeonschaub (approved):
Yes, agree this makes sense. The test output is often too much anyways

PR #394, opened 2025-10-22
[KA] support unified memory API

simeonschaub:


maleadt:
`SharedVirtualMemory` is also shared. Maybe we should do something smarter based on the back-end, a la https://github.com/JuliaGPU/OpenCL.jl/blob/53f450a4e063f32b5873da5af98aec9e0f007a3d/src/array.jl#L96-L105 (which could also be used in the mixed broadcast implementation).

simeonschaub:
The issue I saw with that is that if we have no USM support but support for BDA, we'll always pick `BufferBackend` over `SVMBackend`, in that case we won't be able to support unified memory. If you don't think we should care about this edge case I can also simply use `memory_backend` though

maleadt (approved):
Yeah go ahead. I'm not entirely happy with the design, but this is internal anyway.

PR #409, opened 2026-01-03
Add relevant package versions to `versioninfo`

christiangnrd:


christiangnrd:
@simeonschaub I don't know why you got co-author credit but if you're fine with it I won't bother removing it

PR #410, opened 2026-01-08
Auto-detect SPIR-V extensions from device capabilities

SimonDanisch:
This is needed for supporting Atomix in KernelAbstractions, since we can't set the extension from a KA kernel launch.
This seems to be the most minimal fix, not sure if there are any problems with this or if it would need to be more complete.


lib/intrinsics/src/integer.jl
```diff
@@ -36,6 +36,8 @@ for gentype in generic_integer_types
 
 @device_function popcount(x::$gentype) = @builtin_ccall("popcount", $gentype, ($gentype,), x)
 
+@device_override Base.bitreverse(x::$gentype) = @builtin_ccall("bit_reverse", $gentype, ($gentype,), x)
```
simeonschaub, line 39: This seems useful but unrelated, do you want to open a separate PR?

simeonschaub:
Should be superseded by #453

PR #414, opened 2026-01-11
Shady workaround for Windows crashes with USM buffers

christiangnrd:
Split off from #405 since this is somewhat questionable and I don't want it to block the other fixes.


christiangnrd:
Properly fixed in https://github.com/pocl/pocl/pull/2188

PR #423, opened 2026-03-04
AcceleratedKernels accumulate and sort

michel2323:
This adds accumulate and sort through AcceleratedKernels.jl to be on par with CUDA.jl, AMDGPU.jl, and oneAPI.jl.


src/accumulate.jl
```diff
@@ -0,0 +1,21 @@
+# Use a smaller block size to work around a scan correctness issue
+# with the Blelloch parallel prefix sum at larger block sizes (>=128).
+const _ACCUMULATE_BLOCK_SIZE = 64
```
simeonschaub, line 3: Can you expand on this? Is this a general AK issue or specific to OpenCL? Does this not depend on the workgroup/subgroup size?

PR #428, opened 2026-04-15
Override ^(Float, Int64) to avoid Float64 widening

maleadt:
Base.`^(::Union{Float16,Float32,Float64}, ::Int64)` widens Float32/16 through Float64 (via `power_by_squaring`), which breaks on backends without FP64 support like oneAPI on some devices, and leaves a runtime `pown` call in the generated code in all cases.

Needed for https://github.com/JuliaGPU/GPUArrays.jl/pull/707

### DIFF
--- a/lib/intrinsics/src/math.jl
+++ b/lib/intrinsics/src/math.jl
@@ -1,5 +1,7 @@
 # Math Functions
 
+using Base: @assume_effects
+
 # TODO: vector types
 const generic_types = [Float16, Float32, Float64]
 const generic_types_float = [Float32]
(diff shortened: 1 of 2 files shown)


lib/intrinsics/src/math.jl
```diff
@@ -183,6 +185,36 @@ end
...
+    y == 3  && return x * x * x
+    x ^ (y % Int32)
+end
+@device_override @assume_effects :foldable @inline function Base.:(^)(x::Float64, y::Int64)
+    y == -1 && return inv(x)
+    y == 0  && return one(x)
+    y == 1  && return x
+    y == 2  && return x * x
+    y == 3  && return x * x * x
+    x ^ (y % Int32)
```
nhz2, line 215: Shouldn't this normally overflow to Inf instead of wrapping? Maybe replace `y%Int32` with `Float32(y)`?

```julia-repl
julia> 2.0^(Int64(2)^32 % Int32)
1.0

julia> 2.0^(Int64(2)^32)
Inf
```

maleadt: Yeah OK that's sloppy. Thanks; I'll create a follow-up PR.

PR #431, opened 2026-05-12
Switch to CompilerCaching.jl

maleadt:
x-ref https://github.com/JuliaGPU/GPUCompiler.jl/pull/794


maleadt:
The windows failure here is unrelated, and caused by the LLVM 22 upgrade on Yggdrasil. Will investigate separately.

Issue #417, opened 2026-02-04
PoCL: Upstream Julia CI fails


maleadt:
> Hi, we have recently encountered a bunch of failing Julia workflows like [this one](https://github.com/pocl/pocl/actions/runs/21636567643/job/62363642274?pr=2117). They have some things in common, AFAICT it's always Julia 1.10,  the job runs on Zen4 CPU (AMD EPYC 9V74) and fail with `Invalid instruction at <hex>`. I've tried to reproduce locally with AVX-512 machine, but all the tests pass for me, and i can't really spend much time on debugging this. Have you seen this issue ? Is this a Julia 1.10 specific bug related to AVX512 ? Unfortunately i don't think we can limit the CPU in Github-hosted runners, however we could try using a self-hosted runner.

_Originally posted by @franz in https://github.com/pocl/pocl/issues/2117#issuecomment-3842302978_

franz:
It seems my initial assessment was incorrect; it's also happening on [Julia 1.12](https://github.com/pocl/pocl/actions/runs/21839711810/job/63020451523?pr=2119) and [Windows MSVC](https://github.com/pocl/pocl/actions/runs/21839707728?pr=2119) CI runs. The CPU (`znver4`) and Invalid/Illegal instructions remain the common features. I still can't reproduce locally, but since it also happens on Windows CI, this is most likely not a Julia bug.

vchuravy:
Thanks for letting us know! The error is still confusing, since the disassembly seems to be valid and IIUC a stack misalignment should not be an invalid instruction

franz:
FWIW I've now added a workaround in the CI that limits the CPU to AVX2, and it the Julia CI now passes the tests on Znver4 machine.

Issue #388, opened 2025-10-13
AcceleratedKernels.cumsum runs endlessly when using pocl's cpu backend on Intel CPU


VarLad:
I've been able to reproduce it on two Intel devices. Would be nice if someone could test on AMD.
Rusticl works just fine here.

Julia: 1.11.3
OpenCL: 0.10.5
pocl: pocl_jll.jl (v7.0.0+15)

VarLad:
Still reproducible as of POCL 7.1

shreyas-omkar:
### Versions

```
julia 1.12.5
OpenCL 0.10.10
KernelAbstractions 0.9.42
pocl_jll 7.1.3+5
Platform: POCL, CPU device
```

### MWE

```julia
using pocl_jll, OpenCL, KernelAbstractions
import KernelAbstractions as KA

@inline cfo(n) = n >> 0x5   # conflict-free bank offset (log2 banks = 5)

@kernel inbounds=true unsafe_indices=true function mwe!(v, prefixes, inclusive)
    len = length(v)
    @uniform bs = @groupsize()[1]
    temp = @localmem Int32 (0x2 * bs + cfo(0x2 * bs),)
    ib = @index(Group, Linear) - 0x1
    it = @index(Local, Linear) - 0x1
    boff = ib * bs * 0x2
    ai = it; bi = it + bs
    oa = cfo(ai); ob = cfo(bi)
    temp[ai + oa + 0x1] = boff + ai < len ? v[boff + ai + 0x1] : 0x0
    temp[bi + ob + 0x1] = boff + bi < len ? v[boff + bi + 0x1] : 0x0

    off = typeof(it)(1); np2 = bs * 0x2
    d = np2 >> 0x1
    while d > 0x0                              # up-sweep
        @synchronize()
        if it < d
            _a = off * (0x2 * it + 0x1) - 0x1; _a += cfo(_a)
            _b = off * (0x2 * it + 0x2) - 0x1; _b += cfo(_b)
            temp[_b + 0x1] = temp[_b + 0x1] + temp[_a + 0x1]
        end
        off = off << 0x1; d = d >> 0x1
    end
    it == 0x0 && (temp[np2 - 0x1 + cfo(np2 - 0x1) + 0x1] = 0x0)
    d = typeof(it)(1)
    while d < np2                              # down-sweep
        off = off >> 0x1
        @synchronize()
        if it < d
[...]

PR #439, opened 2026-06-11
Fix CI failures: require fixed PoCL and drop Intel macOS

maleadt:
CI has been red on main for a while. Investigation of the recent runs shows two independent issues:

## Windows: access violation in `pocl_reset_indirect_ptrs` (#317)

USM jobs on Windows crashed in `pocl_reset_indirect_ptrs` (e.g. [this job](https://github.com/JuliaGPU/OpenCL.jl/actions/runs/26944522358/job/79494039212)). This was fixed upstream (pocl/pocl#2188, backported in JuliaPackaging/Yggdrasil#13896) and shipped in pocl_jll 7.1.2; runs since June 4 no longer exhibit the crash. This PR requires `pocl_jll = "7.1.2"` in the test environment so the resolver can't pick known-bad builds again.

## Intel macOS + Julia 1.10: jobs hang until timeout

All `macOS-26-intel` + Julia 1.10 jobs (9/9 over the last three runs on main/PRs) hung until the 100-minute timeout. Log analysis shows the runner kept making progress: in every job, one worker started a testset that never finished while the others completed the rest of the suite. The hung testset was different every time (`gpuarrays/linalg/norm`, `intrinsics`, `device/random`, `gpuarrays/statistics`, `kernelabstractions`, ...), so this is not a bug in a specific test.

This is not fixed by JuliaGPU/GPUCompiler.jl#812: the hanging jobs already ran GPUCompiler v1.20.0, which includes that change (it shipped in v1.13.3), and the local reproducer below still hangs with v1.21.0.

The hang is reproducible locally on Linux (AMD Zen 5) with Julia 1.10 using the MWE from #388:

```julia
using OpenCL, AcceleratedKernels, pocl_jll
x = CLArray([1, 2, 3, 4, 5])
y = AcceleratedKernels.cumsum(x)  # spins forever at 100% CPU inside the kernel
```

The host side just blocks waiting on the device-to-host copy while a PoCL worker thread spins inside the kernel, i.e. PoCL's default work-group lowering (`loopvec`) miscompiles barrier-heavy kernels into infinite loops on some CPUs (and produces silently incorrect results at other block sizes; see #388 and pocl/pocl#1971 — also reminiscent of the Julia 1.10-specific crashes PoCL's upstream CI sees on Zen 4, #417).

This PR therefore:
- drops Intel macOS from the test matrix altogether (GitHub is phasing those runners out anyway);
- sets `POCL_WORK_GROUP_METHOD=cbs` for the remaining test jobs: the miscompilation isn't macOS-specific (reproduced on Linux/Zen 5), and GitHub-hosted runners occasionally come with affected CPUs (PoCL's upstream CI hits this on GitHub-hosted Zen 4 runners, #417).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         version: ['1.10', '1.12']
-        os: [ubuntu-24.04, ubuntu-24.04-arm, macOS-26-intel, windows-2022]
+        os: [ubuntu-24.04, ubuntu-24.04-arm, windows-2022]
         arch: [x64, arm64]
         pocl: [jll, local]
         memory_backend: [usm, svm, buffer]
@@ -35,12 +35,7 @@ jobs:
             arch: arm64
           - os: ubuntu-24.04-arm
             arch: x64
-          # macOS-26-intel is Intel-only
-          - os: macOS-26-intel
-            arch: arm64
           # we only test building PoCL on Linux
-          - os: macOS-26-intel
-            pocl: local
           - os: windows-2022
             pocl: local
     steps:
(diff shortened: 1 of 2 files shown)


maleadt:
Oh oops forgot to mark this as draft; I intended to still clean this up. Anyway, not a big issue.

christiangnrd:
> Oh oops forgot to mark this as draft; I intended to still clean this up. Anyway, not a big issue.

Sorry! Green CI in OpenCL.jl was just too enticing I was less cautious than usual

PR #444, opened 2026-06-17
Only set sub_group_size by default when supported

christiangnrd:
Follow-up to #442


christiangnrd:
Merging since important and we can fix style issues after if people have a nice solution

PR #447, opened 2026-06-24
Add compile-time device feature queries

maleadt:
Register the optional device features we care about (fp16, fp64, int64_atomics, subgroups, generic_address_space) and query them two ways: `device_features(dev)` host-side, and `has_feature(name)` device-side, which folds to a constant so a kernel can pick a path at compile time.

Use it to give the device RNG a per-work-item fallback when subgroups aren't available, and gate the direct subgroup tests on device support.

### DIFF
--- a/lib/cl/kernel.jl
+++ b/lib/cl/kernel.jl
@@ -202,15 +202,22 @@ function enqueue_kernel(k::Kernel, global_work_size, local_work_size=nothing;
     end
 
     if rng_state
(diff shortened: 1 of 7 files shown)


simeonschaub:
I imagine it might be more involved, but could we have a test that the query is actually optimized away? Otherwise looks great!

maleadt:
Added a test.

simeonschaub (approved):

PR #453, opened 2026-07-01
Add floating-point atomic add/sub and min/max intrinsics

michel2323:
Float32 atomic_add!/atomic_sub! emit OpenCL-style builtins that lower to the SPV_EXT_shader_atomic_float_add EXT instructions. Float64 atomic add, and float min/max for both types, use cmpxchg fallbacks since the corresponding native extensions aren't available in the oneAPI set.


michel2323:
The errors come from https://github.com/JuliaGPU/OpenCL.jl/pull/452

simeonschaub:
Could we use an approach like in #447 to query whether the device supports SPV_EXT_shader_atomic_float_add and use the fallback cmpxchg loop only when necessary?

michel2323:
I registered fp32/fp64 atomic_add and atomic_min_max features, detected through the cl_ext_float_atomics capability bitfields (requiring both the global- and local-memory bits, since the intrinsics cover both address spaces). SPIRVIntrinsics now carries a _native! and a _fallback! version of each op, with the public functions defaulting to the compare-and-swap loop and OpenCL.jl overriding them to pick the native instruction via `has_feature`, which folds at compile time. The compiler config also enables the corresponding SPIR-V extensions automatically on devices that report support, so no explicit extensions kwarg is needed anymore.

Making the selection device-driven shook out a few issues, split into separate commits:

- The OpenCL-style builtin names turned out to lower only on the LLVM SPIR-V backend for atomic_add, and on neither backend for min/max — hard link failures on pocl, which advertises cl_ext_float_atomics. The native atomics are now emitted as `__spirv_AtomicF{Add,Min,Max}EXT` wrapper builtins like the integer atomics, which both backends handle. Net effect: Float64 add, and float min/max also get native instructions where supported, and Float32 add/sub gains a fallback where not.
- `spirv2clc` doesn't implement the float min/max capabilities, so those features are masked on the OpenCL C source backend, and kernels take the CAS fallback there.
- The CAS fallback loops now compare raw bits instead of float values, because `==` spins forever on a stored NaN and mistakes a failed exchange for success when `-0.0 == 0.0`.
- `@atomic` float +/-/min/max now route through these intrinsics instead of the generic value-comparing CAS loop.

simeonschaub:
[`SPV_EXT_shader_atomic_float_min_max`](https://github.khronos.org/SPIRV-Registry/extensions/EXT/SPV_EXT_shader_atomic_float_min_max.html) at least also has support for fp16, and for atomic add there is [`SPV_EXT_shader_atomic_float16_add`](https://github.khronos.org/SPIRV-Registry/extensions/EXT/SPV_EXT_shader_atomic_float16_add.html). Should we add those as well while we're at it?

Overall, this looks great to me! Pinging @vchuravy regarding what the story for `KA.PoCLBackend` should look like here, since we probably want to use the native versions wherever possible as well there, ideally without duplicating all the logic in OpenCL.jl

lib/intrinsics/src/atomic.jl
```diff
@@ -84,6 +84,80 @@ end
...
+@device_function atomic_max_native!(p::LLVMPtr{$gentype,$as}, val::$gentype) =
+    @builtin_ccall("__spirv_AtomicFMaxEXT", $gentype,
+                   (LLVMPtr{$gentype,$as}, UInt32, UInt32, $gentype),
+                   p, UInt32(atomic_scope),
+                   UInt32(atomic_memory_semantics(Val($as))), val)
+
+end
+
+# the loops compare raw bits, not values: `==` on floats would spin forever on a stored NaN,
+# and would treat a failed exchange as successful when -0.0 compares equal to 0.0.
```
simeonschaub: Can we not just use `===` for the comparison instead of `==`?

michel2323:
@simeonschaub Sorry for the late reply. I currently have many moving parts in oneAPI.jl. This adds `===` and fp16 support. I only test compilation for fp16 since pocl doesn't support it.

simeonschaub:
> I only test compilation for fp16 since pocl doesn't support it.

Ah, I thought it had them, but you are right!

lib/intrinsics/src/atomic.jl
```diff
@@ -107,6 +180,41 @@ for as in atomic_memory_types
...
+    valbits = UInt32(reinterpret(UInt16, val)) << shift
+    word = Base.unsafe_load(wp, 1)
+    while true
+        (word & mask) == cmpbits ||
+            return reinterpret(Float16, ((word & mask) >> shift) % UInt16)
+        seen = atomic_cmpxchg!(wp, word, (word & ~mask) | valbits)
+        seen === word && return cmp
+        word = seen
+    end
+end
```
simeonschaub: I'm not sure I'm qualified to judge whether this masking trick is correct/advisable, so I'd appreciate if someone else could look over this

christiangnrd:
The pocl next (7.2) has Float16 support

simeonschaub (approved):
Very nice, thank you! Before merging I'd like someone else to take a look at the masking workaround for fp16 atomic_cmpxchg!, but otherwise LGTM

simeonschaub:
> The pocl next (7.2) has Float16 support

Does it? I get:

```julia-repl
julia> using OpenCL, pocl_next_jll

julia> OpenCL.has_fp_atomics(cl.device(), cl.CL_DEVICE_HALF_FP_ATOMIC_CAPABILITIES_EXT, cl.CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT)
false
```

(I think it has fp16 support, but not fp16 atomics)

christiangnrd:
> (I think it has fp16 support, but not fp16 atomics)

Oh shoot you're right. I misunderstood sorry for the noise!

michel2323:
I am also a bit uncertain about that mask workaround. It's LLM-generated, and I just found a bug in it. Maybe we should just remove FP16 `atomic_cmpxchg!` support for now until we can execute that path with the LLVM backend on some hardware.

simeonschaub:
Yes, that seems like a good idea!

michel2323:
I've removed the fp16 atomics. I have one path in oneAPI.jl that should generate fp16 atomics with the Khronos SPIRV translator, but that's for one system (Aurora) and thus a bit exotic. I'll revisit once that path also moves to the LLVM SPIRV translator.

simeonschaub (approved):
LGTM. I'll merge tomorrow if there are no objections

PR #456, opened 2026-07-09
Trim the CI matrix to hand-picked configurations.

### LINKED ISSUE #366: Fix and slim down CI

maleadt:
Instead of crossing all memory backends with all platforms, Julia versions, and PoCL builds (33 jobs, ~1000 runner-minutes), run 10 hand-picked configurations: the memory backend is a host-side, load-time preference that is essentially orthogonal to the other axes, so cover all backends on Linux x64 and rotate them across the other platforms. The from-source PoCL build (upstream HEAD) is reduced to a single job, and the separate pocl_next smoke-test job is folded into the main matrix as a `pocl: next` axis value, now covering every platform rather than just Linux x64.

Closes https://github.com/JuliaGPU/OpenCL.jl/issues/366

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -21,23 +21,30 @@ jobs:
       contents: read
     strategy:
       fail-fast: false
+      # Hand-picked configurations instead of a full cross-product: the memory backend
+      # is a host-side, load-time preference that's essentially orthogonal to OS, arch,
+      # Julia version and PoCL build, so we cover all backends on the primary platform
+      # (Linux x64) and rotate them across the other platforms. The `pocl` axis selects
+      # the released JLL (`jll`), the upcoming release (`next`, via pocl_next_jll, run
+      # on every platform), or upstream PoCL HEAD built from source (`local`, one job,
+      # Linux-only).
       matrix:
-        version: ['1.10', '1.12']
-        os: [ubuntu-24.04, ubuntu-24.04-arm, windows-2022]
-        arch: [x64, arm64]
-        pocl: [jll, local]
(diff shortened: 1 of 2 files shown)


christiangnrd:
This is good. Can we do 1.11 and 1.13 too?

.github/workflows/Test.yml
```diff
@@ -134,7 +144,7 @@ jobs:
         uses: julia-actions/julia-runtest@v1
         if: runner.os != 'Windows'
         with:
-          test_args: '--quickfail --verbose --platform=pocl'
+          test_args: "--quickfail --verbose --platform=${{ matrix.pocl == 'next' && 'pocl_next' || 'pocl' }}"
```
simeonschaub:
```suggestion
          test_args: "--quickfail --verbose --platform=${{ case(matrix.pocl == 'next', 'pocl_next', 'pocl') }}"
```
might be a more natural way to spell this?

simeonschaub (approved):
Thanks! I think this makes a lot of sense to save some unnecessary CI hours

PR #457, opened 2026-07-14
handle missing vendor info in OpenCL version string gracefully

simeonschaub:
On my laptop, XRT advertises the NPU via OpenCL, but the version returned by the query is just `"OpenCL 1.0"` without any vendor info. Handle that case gracefully so at least `OpenCL.versioninfo()` doesn't error


christiangnrd:
Can you rebase?

christiangnrd:
https://github.com/JuliaGPU/OpenCL.jl/actions/runs/29565903596/job/87838349478?pr=457 seems real but unrelated

PR #458, opened 2026-07-14
Make the translator configurable and tested

simeonschaub:
Requires https://github.com/JuliaPackaging/Yggdrasil/pull/14212 and
https://github.com/JuliaGPU/GPUCompiler.jl/pull/879 to pass.

Deletes one float atomics test I added before we had proper support for
float atomics and which is not really relevant anymore. (ref
https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3404)

### DIFF
--- a/.github/workflows/Test.yml
+++ b/.github/workflows/Test.yml
@@ -13,7 +13,7 @@ concurrency:
 
 jobs:
   test:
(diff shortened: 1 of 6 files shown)


simeonschaub:
> Deletes one float atomics test I added before we had proper support for
> float atomics and which is not really relevant anymore.

Just realized I was thinking of #453, which isn't merged yet, so we might want to merge that PR first

maleadt (approved):

Issue #422, opened 2026-02-21
LLVM SPIR-V backend crashes on functions returning small static arrays (e.g. SMatrix{1,1,Float32})


ChrisRackauckas-Claude:
## Summary

The LLVM SPIR-V backend (`llc` from `SPIRV_LLVM_Backend_jll`) crashes with `LLVM ERROR: Broken function found, compilation aborted!` when compiling GPU kernels that call functions returning small static array types like `SMatrix{1,1,Float32}`.

The root cause is that `SMatrix{1,1,Float32}` gets lowered to LLVM type `[1 x float]`, and the SPIR-V backend incorrectly transforms the return type to `i32` during internal passes, causing the LLVM IR verifier to abort.

## MWE

```julia
using OpenCL, pocl_jll, KernelAbstractions, StaticArrays

# A kernel that calls a function returning SMatrix{1,1,Float32}
# This gets lowered to `[1 x float]` in LLVM IR, which the SPIR-V backend miscompiles
@kernel function smatrix_return_kernel(out)
    i = @index(Global)
    A = SMatrix{1,1,Float32}(3.0f0)
    B = SMatrix{1,1,Float32}(2.0f0)
    C = A * B  # returns SMatrix{1,1,Float32}
    out[i] = C[1]
end

backend = OpenCL.OpenCLBackend()
out = KernelAbstractions.zeros(backend, Float32, 4)
kernel = smatrix_return_kernel(backend, 4)
kernel(out, ndrange=4)
KernelAbstractions.synchronize(backend)
```

## Error output

```
LLVM ERROR: Broken function found, compilation aborted!
ERROR: LoadError: Failed to translate LLVM code to SPIR-V.
...
caused by: failed process: Process(`...llc ...`, ProcessSignaled(6)) [0]
```

## Context

This was discovered via [SciML/DiffEqGPU.jl](https://github.com/SciML/DiffEqGPU.jl), where stiff ODE solvers using `autodiff=false` call `finite_diff_jac`, which returns `SMatrix` types. The `finite_diff.jl` tests fail on all Julia versions (lts, 1, pre) when using the OpenCL backend. See https://github.com/SciML/DiffEqGPU.jl/actions/runs/22247943080/job/64365791848

[...]

PR #459, opened 2026-07-15
Add aggregate return regression test

maleadt:
Test for https://github.com/JuliaGPU/OpenCL.jl/issues/422

### DIFF
--- a/test/kernelabstractions.jl
+++ b/test/kernelabstractions.jl
@@ -1,6 +1,24 @@
 # KernelAbstractions has a testsuite that isn't part of the main package.
 # Include it directly.
 
+using KernelAbstractions, StaticArrays
+
+@testset "#422: SMatrix return" begin
+    @kernel function smatrix_return_kernel(out)
+        i = @index(Global)
+        A = SMatrix{1,1,Float32}(3.0f0)
+        B = SMatrix{1,1,Float32}(2.0f0)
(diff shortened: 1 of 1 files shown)


maleadt:
@simeonschaub No need to use the translator now :-)

simeonschaub:
Oh, great! I think there is still some value in having options, especially if we can help upstream identify some of the edge cases, but maybe that's just me :).

Should we add the full Julia-side reproducer as an integration test as well?

maleadt:
Yeah probably better than the current hacky IL test.

MrSidims:
Sorry for a bit lurking around :)

> Oh, great! I think there is still some value in having options, especially if we can help upstream identify some of the edge cases, but maybe that's just me :).

There are some edge cases, but overall I'd consider SPIR-V backend as production ready. At least we have switched it to be default SPIR-V producer for HIP amdgcnspirv target. Two known caveats:
1. debug info support is not fully implemented in the backend, that it's why we kept -no-use-spirv-backend option in the frontend for users, who care about debug info;
2. currently no optimizations are enforced pre-SPIR-V generation; if enforced, actually there won't be many issues, just few HeCBench samples would start failing, I'll open LLVM issues one day or another, when restore my logs and re-check pass rates, but I have a feeling that your fix for aggregate type intrinsics would fix half of them.

PR #461, opened 2026-07-16
Update version to 0.10.10

christiangnrd:
Was released in a separate branch.

Add verbose because test durations and listed runtimes are a bit suspicious


christiangnrd:
Failure is because https://github.com/pocl/pocl/pull/2235 isn't merged upstream yet

PR #470, opened 2026-08-06
Bump version

christiangnrd:


simeonschaub (approved):
(I was waiting to merge #458 before I made the release, but no need to hold this up really)

christiangnrd:
This was to try and get the AcceleratedKernels CI running the latest version of all backends but it seems like we're going in the direction of changing the AK CI to be less restrictive so I'll let you decide if we merge now or wait for #458

PR #471, opened 2026-08-07
Test windows-2025

christiangnrd:
Some mingw fixes may have fixed the issues we had with windows-2025 runners


christiangnrd:
Not fixed

PR #472, opened 2026-08-11
Thread safety fixes

maleadt:
Improves thread safety by locking some operations. Annoyingly, `Managed` also needs to become transactional to protect against:

```text
Initial state: managed.queue = Q0, managed.dirty = true

Task A                                      Task B
------                                      ------
convert for Q1
  finish(Q0)
  managed.queue = Q1
  managed.dirty = true
return pointer
                                            convert for Q2
                                              finish(Q1)
                                              # Q1 is empty at this instant
                                              managed.queue = Q2
                                              managed.dirty = true
                                            return pointer
                                            enqueue operation B on Q2
enqueue operation A on Q1
```

`finish(Q1)` does not order operation A because A has not been submitted yet. Operations A
and B can consequently execute concurrently even though the `Managed` state claims that
ownership was transferred from Q1 to Q2. Later synchronization consults only the one queue
stored in `managed.queue`; it cannot reconstruct the missing ordering edge.

### DIFF
--- a/lib/cl/kernel.jl
+++ b/lib/cl/kernel.jl
@@ -7,14 +7,19 @@ export clcall
 
 mutable struct Kernel <: CLObject
     const id::cl_kernel
+    const lock::ReentrantLock
 
     function Kernel(k::cl_kernel, retain::Bool=false)
-        kernel = new(k)
+        kernel = new(k, ReentrantLock())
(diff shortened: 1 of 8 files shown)


maleadt:
@simeonschaub Now cleaner :-)

PR #478, opened 2026-09-04
Report device-side exceptions at host synchronization

maleadt:
OpenCL kernels can now report failures to the host as `KernelException`s. A faulting
work-item returns, and the error is raised when the host synchronizes, including through
`cl.finish`, blocking copies such as `Array(a)`, and KernelAbstractions synchronization.
The report is consumed before throwing, so subsequent kernels can run normally.

```julia
using OpenCL

a = OpenCL.zeros(Float32, 1)
@opencl debug_level=2 (a -> (a[2] = 1f0; return))(a)
cl.finish(cl.queue())  # KernelException with BoundsError details and a device backtrace
```

The new `debug_level` keyword defaults to Julia's `-g` setting. Level 0 records only that
an exception occurred; level 1 adds names and reasons for common runtime errors; level 2
also records work-item coordinates and a bounded device-side backtrace.

Each device in a context has a mailbox shared by its queues. Synchronizing one queue may
therefore wait for and report an error from another queue on that device. Device atomics
select one writer, identified by its launch and work-item coordinates, so later kernels
cannot overwrite its diagnostics. Finalizer-driven synchronization leaves reports pending
for ordinary host checks.

Mailbox storage uses USM host memory, fine- or coarse-grained SVM, or a mappable buffer
with a device address. Unsupported contexts warn and run without host-side reporting.
Mailbox allocations retain their contexts for the lifetime of the process. Failed kernels
may leave incomplete results, and work-group barriers must still obey convergence rules.

Requires GPUCompiler 2.5.2 for SPIR-V exception lowering, failed-boxing handling, and the
correct address space for diagnostic strings. This supersedes #328 and the earlier
host-side prototype in #342.

Validated with native Intel and NVIDIA drivers, packaged PoCL 7.1 and 7.2, and a local
PoCL build, using both SPIR-V and OpenCL C paths where supported. Tests cover Julia 1.10,
1.12, and 1.13, debug levels, boxed exceptions, blocking copies, concurrent checks,
report ownership, mailbox memory backends, and finalizer synchronization. Documentation
builds successfully.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -31,7 +31,7 @@ SPIRVIntrinsics = {path = "lib/intrinsics"}
 [compat]
 Adapt = "4"
 GPUArrays = "11.2.1"
-GPUCompiler = "2"
+GPUCompiler = "2.5.2"
 KernelAbstractions = "0.9.38"
 LLVM = "9.6"
 LinearAlgebra = "1"
--- a/docs/src/02-exceptions.md
+++ b/docs/src/02-exceptions.md
@@ -1,0 +1,64 @@
+# Device-side exceptions
+
+OpenCL kernels report device-side exceptions to the host as `KernelException`s. The
+faulting work-item exits, and the exception is raised at the next synchronization point:
+`cl.finish`, a blocking copy such as `Array(x)`, or `KernelAbstractions.synchronize`.
+
+```julia-repl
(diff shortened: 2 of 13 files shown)


maleadt:
CI failure is https://github.com/JuliaGPU/OpenCL.jl/pull/479

PR #483, opened 2026-09-14
Add KernelAbstractions device selection interface

christiangnrd:
The KA 0.10 docs (rightfully) make this a mandatory implementation if a backend supports multiple devices.

This adds a `platform` field to `OpenCLBackend`.


src/OpenCLKernels.jl
```diff
@@ -43,6 +43,43 @@ Adapt.adapt_storage(::OpenCLBackend, a::Array) = Adapt.adapt(CLArray, a)
 Adapt.adapt_storage(::OpenCLBackend, a::CLArray) = a
 Adapt.adapt_storage(::KA.CPU, a::CLArray) = convert(Array, a)
 
+## Device Selection
+
+# devices are numbered consecutively across all platforms, in enumeration order
+
+function KA.ndevices(::OpenCLBackend)
```
vchuravy, line 50: This makes me wonder if the OpenCLBackend should take a platform as an argument.

christiangnrd: I like this a lot. I could keep the within-platform device sorting, get rid of the awful code I (claude) wrote to sort the platforms over in #485, and most functionality isn't meant to work across platforms anyways

vchuravy (approved):

src/OpenCLKernels.jl
vchuravy: Do we now need to check if the platform is the current one, on calls?

christiangnrd: ~We could give `OpenCLBackend` its own context and wrap every call around a `cl.context(backend.ctx) do` block? That might add quite a bit of overhead though~
Fable convinced me this is a bad idea

christiangnrd: @vchuravy I pushed a few fixes/improvements, as well as removing the `platform!` call in the backend contructor, and I added a warning when calling `KA.allocate` or launching a KA kernel with current and backend platform mismatch

src/OpenCLKernels.jl
```diff
@@ -14,10 +14,17 @@ import Adapt
 
 export OpenCLBackend
 
-struct OpenCLBackend <: KA.GPU
+Base.@kwdef struct OpenCLBackend <: KA.GPU
+    platform::cl.Platform = cl.platform()
 end
 
-function KA.allocate(::OpenCLBackend, ::Type{T}, dims::Tuple; unified::Bool = false) where T
+function check_platform(b::OpenCLBackend)
```
vchuravy: You likely want to outline the warning and inline the check.

christiangnrd: If https://github.com/JuliaGPU/OpenCL.jl/pull/483/commits/7f6383e67fcd2abebb66cc16bedd48f2ee130f0c looks good to you, I’ll merge once tests pass

PR #487, opened 2026-09-15
[CI] Try to reproduce local bug in CI

christiangnrd:
On my linux desktop, running the testsuite would crash before any tests ran, which Fable attributed to a race-condition related to JULIA_NUM_THREADS=auto and pocl replacing host signal handlers during initialization?

<details>
<summary>Claude slop:</summary>

# PoCL replaces the host's signal handlers during device initialization

## Summary

Initializing the CPU device (`clGetDeviceIDs`) installs LLVM's process-wide crash
handlers (`SignalHandler` from `llvm/lib/Support/Unix/Signals.inc`) over whatever the
host application had installed. Hosts that rely on their own SIGSEGV handler break.
Julia is one: it implements GC safepoints as a read fault on a protected page, so once
LLVM's handler owns SIGSEGV, the next safepoint hit by any thread terminates the
process. No kernel has been compiled at that point.

## Environment

- PoCL 7.1.3, built against LLVM 20.1.2 (Julia's `pocl_jll` 7.1.3+9), CPU device
  only, x86_64 Linux
- Host: Julia 1.13.0 running with 16 threads

## Steps to reproduce

1. In a host process that has its own SIGSEGV handler, record
   `sigaction(SIGSEGV, NULL, &old)`.
2. Call `clGetPlatformIDs`, then `clGetDeviceIDs` on the PoCL platform.
3. Query `sigaction(SIGSEGV, ...)` again.

Observed: `sa_handler` now points into `libpocl.so` (LLVM's `SignalHandler(int)`).
It is unchanged after `clGetPlatformIDs` and only changes at `clGetDeviceIDs`.

## Evidence

Core dump of the crashed host, faulting thread:

```
#0  SignalHandler(int) () from libpocl.so
#1  <signal handler called>
#2  _jl_mutex_wait (..., safepoint=1) at threading.c:986
#3  _jl_mutex_lock ...
```

The fault at #2 is the host's intentional safepoint fault. Julia's handler would have
parked the thread; LLVM's handler runs its callbacks, restores SIG_DFL, and re-raises,
killing the process with no output. The crash only appears when the host is
multithreaded, since a safepoint must be hit while another thread holds a lock.

## Expected

A library should not install process-wide signal handlers, or should at least make it
opt-out. `POCL_SIGUSR2_HANDLER` already exists for PoCL's own handler; nothing covers
LLVM's. Something in the CPU device's LLVM initialization is calling one of
`PrintStackTraceOnErrorSignal`, `RemoveFileOnSignal`, `AddSignalHandler`,
`SetInterruptFunction`, or `CrashRecoveryContext::Enable`, all of which trigger LLVM's
one-time `RegisterHandlers`. I could not identify which from the binary alone.

## Workaround

The host snapshots its handlers before the first OpenCL call and reinstalls them after
`clGetDeviceIDs` and `clBuildProgram`. Since LLVM registers only once per process, the
reinstalled handlers then stay in place.

</details>


christiangnrd:
Caught in the CUDA Buildkite run: https://buildkite.com/julialang/opencl-dot-jl/builds/1246/table?jid=01a0a55b-5e1c-4196-9d87-3440f5c2f844&tab=output

christiangnrd:
@maleadt This reproduces most but not every time on x86 linux. The simple workaround is to just not set `JULIA_NUM_THREADS`, but maybe worth looking into?

maleadt:
Slop, but seems like a correct diagnosis. PoCL installs signal handlers to catch SIGFPE, which get confused by Julia's SIGSEGV-based GC mechanism. Same with Clang when going through OpenCL C. I'll take a look.

maleadt:
https://github.com/pocl/pocl/pull/2339 and https://github.com/pocl/pocl/pull/2340 should help for the OpenCL C case, and I'll rebuild with `-DENABLE_SIGFPE_HANDLER=OFF` or so to fix the x86 case.

